PR #19, opened 2021-05-02
Convert Java-like code to Kotlin-like code

ChachyDev:


src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
```diff
@@ -219,8 +218,8 @@ class ChatListener {
                 val lines = (chatGui as AccessorGuiNewChat).chatLines
                 val drawnLines = (chatGui as AccessorGuiNewChat).drawnChatLines
                 var i = 0
-                delay(Duration.microseconds(100))
-                while (!StringUtils.stripControlCodes(lines[i].chatComponent.unformattedText)
+                delay(100)
```
My-Name-Is-Jeff: Isn't delay defaulted to milliseconds instead of microseconds

My-Name-Is-Jeff (approved):

Sychic:
Why are 90% of the changes just adding a 1 to the variable names and changing imports

src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
Sychic: yeah I wanted it to be slightly less than a millisecond so it'd clear the lines faster

Sychic (approved):

src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
ChachyDev: ah yeah, woops i'll fix that now. should be resolved in the latest push now

PR #25, opened 2021-05-14
Implement GitHub Actions

proudmuslim-dev:
`gradlew` shows as modified due to me marking it as an executable.
Edit: This PR was redone and merged [here](https://github.com/Skytils/SkytilsMod/pull/26).


.github/workflows/gradle.yml
```diff
@@ -0,0 +1,16 @@
+name: build-ci
```
proudmuslim-dev, line 1: Could probably have a better name for this but I cba to come up with one

proudmuslim-dev:
Let me fix it, forgot to add `setupDecompWorkspace`

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Java setup
+      uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - name: Build
+      run: ./gradlew setupDecompWorkspace && ./gradlew build
```
My-Name-Is-Jeff: Why not setupCiWorkspace?

My-Name-Is-Jeff: It doesn't seem to upload artifacts either

My-Name-Is-Jeff (changes requested):

.github/workflows/gradle.yml
proudmuslim-dev: You're right, I'd forgotten

proudmuslim-dev: Resolved

proudmuslim-dev: I can fix that as well, one moment

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Java setup
+      uses: actions/setup-java@v1
+      with:
+        java-version: 8
```
My-Name-Is-Jeff: Is it possible to add the cache-packages task

My-Name-Is-Jeff (changes requested):

proudmuslim-dev:
This will be reopened once it isn't a dumpster fire and I can actually focus

PR #26, opened 2021-05-15
Implement GitHub Actions

proudmuslim-dev:
No longer a dumpster fire


My-Name-Is-Jeff (approved):

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,30 @@
...
+          path: build/libs
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
```
DJtheRedstoner, line 23: Caching should be the first action, not the last action. The cache action restores from the cache and registers a "clean up" action that puts into the cache.

DJtheRedstoner (changes requested):

.github/workflows/gradle.yml
proudmuslim-dev: Resolved

PR #31, opened 2021-06-03
Add options to hide sea creature catch and fishing treasure messages

LeaPhant:
_bear with me I haven't written a single line of Kotlin in my life_

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -2220,7 +2220,7 @@ class Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sorting
     @Property(
         type = PropertyType.SELECTOR,
         name = "Legendary Sea Creature Catch Hider",
-        description = "Removes legendary sea creature catch messages as well.",
+        description = "Removes legendary sea creature catch messages from fishing.",
         category = "Spam",
         subcategory = "Fishing",
         options = ["Normal", "Hidden", "Separate GUI"]
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -2207,6 +2207,36 @@ class Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sorting
     )
     var blessedBaitHider = 0
 
+    @Property(
+        type = PropertyType.SELECTOR,
+        name = "Sea Creature Catch Hider",
+        description = "Removes regular sea creature catch messages from fishing.",
+        category = "Spam",
+        subcategory = "Fishing",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
+    var scCatchHider = 0
+
+    @Property(
+        type = PropertyType.SELECTOR,
+        name = "Legendary Sea Creature Catch Hider",
+        description = "Removes legendary sea creature catch messages as well.",
+        category = "Spam",
+        subcategory = "Fishing",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
+    var legendaryScCatchHider = 0
+
+    @Property(
+        type = PropertyType.SELECTOR,
+        name = "Fishing Treasure Hider",
+        description = "Removes good/great catch messages from fishing.",
+        category = "Spam",
+        subcategory = "Fishing",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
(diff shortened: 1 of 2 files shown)


src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
```diff
@@ -69,6 +69,43 @@ class SpamHider {
     private var abilityUses = 0
     private var lastAbilityUsed = ""
 
+    private var SEA_CREATURES = setOf<String>(
```
DJtheRedstoner: This should be a val (means its final/constant), not var and the explicit generic is not required.

src/main/kotlin/skytils/skytilsmod/core/Config.kt
```diff
@@ -2207,6 +2207,46 @@ class Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sorting
...
+        category = "Spam",
+        subcategory = "Fishing",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
+    var scCatchHider = 0
+
+    @Property(
+        type = PropertyType.SELECTOR,
+        name = "Legendary Sea Creature Catch Hider",
+        description = "Removes legendary sea creature catch messages as well.",
```
DJtheRedstoner: "as well" sounds a bit informal to me.

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
```diff
@@ -69,6 +69,43 @@ class SpamHider {
...
+        "You reeled in a Sea Archer.",
+        "The Monster of The Deep emerges from the dark depths...",
+        "Huh? A Catfish!",
+        "Is this even a Fish? It's the Carrot King!",
+        "Gross! A Sea Leech!",
+        "You've discovered a Guardian Defender of the sea.",
+        "You have awoken the Deep Sea Protector, prepare for a battle!",
+    )
+
+    private var LEGENDARY_SEA_CREATURES = setOf<String>(
```
DJtheRedstoner: Same as above.

DJtheRedstoner (changes requested):

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
```diff
@@ -380,6 +417,42 @@ class SpamHider {
...
+            // Legendary sea creature catch
+            if (LEGENDARY_SEA_CREATURES.contains(unformatted)) {
+                when (Skytils.config.legendaryScCatchHider) {
+                    1, 2 -> cancelChatPacket(event, Skytils.config.legendaryScCatchHider == 2)
+                    else -> {
+                    }
+                }
+            }
+
+            // Good catch
```
Sychic, line 438: Could be changed to use formatted but doesn't really matter.

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
```diff
@@ -380,6 +417,42 @@ class SpamHider {
...
+            if (unformatted.startsWith("GOOD CATCH! You found")) {
+                when (Skytils.config.goodTreasureHider) {
+                    1, 2 -> cancelChatPacket(event, Skytils.config.goodTreasureHider == 2)
+                    else -> {
+                    }
+                }
+            }
+
+            // Great catch
```
Sychic, line 447: This one too.

Sychic (approved):

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
My-Name-Is-Jeff: Should be fine because it is impossible for another player to send a message that starts with this

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
My-Name-Is-Jeff: The only issue with using unformatted + contains is that it can trigger on player chat, even though it's unlikely

LeaPhant: Since contains checks against the set in this case it should only trigger if the full chat message is exactly that phrase which a player wouldn't be able to send.

PR #32, opened 2021-06-06
Add multiple new features surrounding Diana event

cuddysteadman:
# Adds multiple new features:
 - Fully integrates "secret command" (/st swaphub) to not be bannable on hypixel - detects mouse input after /warpforge and cancels it (as far as I know should be ok?)
 - Specially colour + nametag the closest burrow (detects by going through all already detected burrows and comparing distance to player)
 - Adds 4 options for the "Looking for burrows... Skytils loaded x burrows!"
      - Toast
      - Separate GUI
      - Normal
      - Hidden
 - Adds a slider for time between burrow refreshes (minimum 30s to not anger API), reflects in gui element
 - Adds an option to track griffin burrow drops based on rarity - the tracker only shows what you can get with the griffin you have atm (legendary, common, none) (requires opening pets menu for detection)


src/main/kotlin/skytils/skytilsmod/core/Config.kt
```diff
@@ -816,6 +816,26 @@ class Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sorting
     )
     var showGriffinCountdown = false
 
+    @Property(
+        type = PropertyType.SLIDER,
+        name = "Interval Between Burrow Refreshes",
+        description = "Amount of time to wait in between griffin burrow API refreshes.",
```
My-Name-Is-Jeff: There is no reason to allow the user to shorten the refresh cooldown, as it has no effect

src/main/kotlin/skytils/skytilsmod/features/impl/events/GriffinBurrows.kt
```diff
@@ -447,11 +473,18 @@ class GriffinBurrows {
                 }
                 return "$type §bPosition: ${chain + 1}/4${
                     if (closest != null) " ${closest.nameWithColor}" else ""
-                }"
+                }${if (getClosestBurrow()!!.equals(this)) " (CLOSEST)" else ""}"
```
My-Name-Is-Jeff: you can simplify this and not use .equals

src/main/kotlin/skytils/skytilsmod/features/impl/trackers/MythologicalTracker.kt
```diff
@@ -123,11 +129,11 @@ class MythologicalTracker : PersistentSave(File(File(Skytils.modDir, "trackers")
             if (event.entity.name == "Minos Champion") {
                 println("Dug is: Minos Champion")
                 lastMinosChamp = 0L
-                BurrowMob.CHAMP.dugTimes++
```
My-Name-Is-Jeff: Unless Hypixel fixed this issue this should not be removed

src/main/kotlin/skytils/skytilsmod/features/impl/trackers/MythologicalTracker.kt
```diff
@@ -278,6 +284,21 @@ class MythologicalTracker : PersistentSave(File(File(Skytils.modDir, "trackers")
     }
 
     class MythologicalTrackerElement : GuiElement("Mythological Tracker", FloatPair(150, 120)) {
+        fun getTotal(): Int {
```
My-Name-Is-Jeff: This is unnecessary, use a filter function

src/main/kotlin/skytils/skytilsmod/features/impl/trackers/MythologicalTracker.kt
```diff
@@ -377,4 +412,27 @@ class MythologicalTracker : PersistentSave(File(File(Skytils.modDir, "trackers")
...
+            if (inventory.lowerChestInventory.displayName.unformattedText.equals("Pets")) { // pets menu
+                var slots = chest.inventorySlots.inventorySlots
+                slots.forEach {
+                    if (it.hasStack) {
+                        if (ItemUtil.isPet(it.stack)) {
+                            val name = it.stack.displayName
+                            if (name.contains("Griffin")) {
+                                Skytils.config.griffinPetRarity = ItemUtil.getRarity(it.stack)!!.rarityInt // set rarity for use above, 0-indexed with -1 as no griffin
+                                Skytils.config.markDirty()
+                                Skytils.config.writeData()
```
My-Name-Is-Jeff: You should not be calling writeData

src/main/kotlin/skytils/skytilsmod/utils/ItemRarity.kt
```diff
@@ -26,28 +26,28 @@ import java.awt.Color
  * https://github.com/SteveKunG/SkyBlockcatia/blob/1.8.9/LICENSE.md
  * @author SteveKunG
  */
-enum class ItemRarity(val rarityName: String, val baseColor: EnumChatFormatting, val color: Color) {
```
My-Name-Is-Jeff, line 29: not really sure what this "rarityInt" is supposed to be for

My-Name-Is-Jeff (changes requested):

src/main/kotlin/skytils/skytilsmod/utils/ItemRarity.kt
cuddysteadman: It's just a way to easily access the rarity without having to use .equals() all the time when comparing rarities (without it you would have to use itemRarity.rarityName.equals(blah blah blah) but instead you can have itemRarity.rarityInt == blah. It's also useful for using operators to compare what should be displayed on the griffin tracker, because with Strings there's no way to compare the relative rarity of items. If there's a better way to do it that I'm unaware of then that's fine with me ¯\_(ツ)_/¯

My-Name-Is-Jeff: you can't do ItemRarity == ItemRarity?

cuddysteadman: Still, how would one compare values with operators?

src/main/kotlin/skytils/skytilsmod/core/Config.kt
cuddysteadman: Oh does the API only refresh it only 60s, or something similar to that? I don't really know how the griffin API works :P

PR #33, opened 2021-06-06
Fix some sea creature spawn messages, optimize detection

LeaPhant:


src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
```diff
@@ -411,10 +411,10 @@ class SpamHider {
                     else -> {
                     }
                 }
-            }
+            }<
```
My-Name-Is-Jeff: Did you accidentally type something here

My-Name-Is-Jeff (changes requested):

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
LeaPhant: Yeah I didn't mean to already PR earlier apologies for the confusion. Fixed it and was able to test the changes now, should be all good.

PR #36, opened 2021-07-04
Remove sychicpet and the skin replacing

TymanWasTaken:
Because many people don't like it and it's just annoying to force the sychic on to random people using the mod

I might have missed some things but whatever


TymanWasTaken:
😡 at least make it a config option

PR #51, opened 2021-07-17
[WIP] Fix TickTask not being registered correctly

Tedyst:
This caused a huge amount of logs being dumped onto the console, causing the main thread to hang in excess of 5 seconds. Based on the limited testing that I did, the issue was not fixed by this commit, but it helped mitigate.


Tedyst:
image
An example of the issue (I am putting this here bc I can't find an issue tracker)

PR #59, opened 2021-08-04
make essential loading less complicated

Wyvest:
its literally impossible for this to go wrong unless it does


Sychic:
Already fixed

PR #60, opened 2021-08-07
Add hide tooltips on storage

AzuredBlue:
also added /skytils update to help


src/main/kotlin/skytils/skytilsmod/commands/SkytilsCommand.kt
```diff
@@ -167,6 +167,7 @@ object SkytilsCommand : CommandBase() {
                             #  §3/skytils setkey §l➡ §bSets your Hypixel API key.
                             #  §3/skytils help §l➡ §bShows this help menu.
                             #  §3/skytils reload <data/mayor/slayer> §l➡ §bForces a refresh of data.
+                            #  §3/skytils update §l➡ §bChecks for updates in-game.
```
My-Name-Is-Jeff, line 170: @Sychic is this supposed to be secret?

Sychic: nope

PR #87, opened 2021-10-17
Use Essential APIs

Wyvest:
Removed Skytils threadPool and replaced with Essential Multithreading
Removed Skytils displayScreen and replaced with Essential's GuiUtils
Rewrite commands to utilize Essential's Command API

**Yes I know that I forked this to W-OVERFLOW instead of my personal GitHub that was on purpose**


My-Name-Is-Jeff:
We will not be using the Essential APIs

Wyvest:
> We will not be using the Essential APIs

Okay

PR #120, opened 2021-12-31
Feature: Add Power Orb Lock minimum Duration

lineargraph:
Power Orb Lock now does not prevent the player from placing another orb if the duration of said orb is below a configurable threshold.

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -2365,6 +2365,8 @@ object Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sortin
         addDependency("boxedTankColor", "boxedTanks")
         addDependency("boxedProtectedTeammatesColor", "boxedProtectedTeammates")
 
+        addDependency("powerOrbDuration", "powerOrbLock")
+
         addDependency("yangGlyphColor", "highlightYangGlyph")
         addDependency("nukekebiHeadColor", "highlightNukekebiHeads")
 
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -1573,6 +1573,14 @@ object Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sortin
     var powerOrbLock = false
 
     @Property(
+        type = PropertyType.NUMBER, name = "Power Orb Lock Duration",
+        description = "Needs Power Orb Lock to be active. Allows overwriting a power orb, if it has less time left than this option.",
+        min = 1, max = 60,
+        category = "Miscellaneous", subcategory = "Quality of Life"
+    )
+    var powerOrbDuration = 10
+
+    @Property(
         type = PropertyType.SWITCH, name = "Prevent Log Spam",
         description = "Prevents your logs from being spammed with exceptions while on Hypixel.",
         category = "Miscellaneous", subcategory = "Quality of Life"
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -1575,7 +1575,7 @@ object Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sortin
     @Property(
         type = PropertyType.NUMBER, name = "Power Orb Lock Duration",
         description = "Needs Power Orb Lock to be active. Allows overwriting a power orb, if it has less time left than this option.",
-        min = 1, max = 60,
+        min = 1, max = 120,
         category = "Miscellaneous", subcategory = "Quality of Life"
     )
     var powerOrbDuration = 10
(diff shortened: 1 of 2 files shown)


lineargraph:
Alright, i only tested with radiant and manaflux cause im poor, but it should be working.

src/main/kotlin/skytils/skytilsmod/core/Config.kt
```diff
@@ -1572,6 +1572,14 @@ object Config : Vigilant(File("./config/skytils/config.toml"), "Skytils", sortin
     )
     var powerOrbLock = false
 
+    @Property(
+        type = PropertyType.NUMBER, name = "Power Orb Lock Duration",
+        description = "Needs Power Orb Lock to be active. Allows overwriting a power orb, if it has less time left than this option.",
+        min = 1, max = 60,
```
My-Name-Is-Jeff: the bubblegum pet item doubles the time but idk if anyone cares

src/main/kotlin/skytils/skytilsmod/features/impl/misc/LockOrb.kt
```diff
@@ -37,30 +39,38 @@ class LockOrb {
...
-                    mc.thePlayer.playSound("random.orb", 0.8f, 1f)
-                    event.isCanceled = true
-                    break
-                }
-            }
+            orbEntity ?: continue
+            val orb = PowerOrbs.getPowerOrbMatchingName(orbEntity.displayName.formattedText) ?: continue
+            if (orb.ordinal < heldOrb.ordinal) continue
+            if (!orb.isPlayerInRange(orbEntity, mc.thePlayer)) continue
+            val remainingDuration = orb.getDurationRemaining(orbEntity)
```
My-Name-Is-Jeff, line 50: ?: continue maybe, merge the above 2 if's idk

src/main/kotlin/skytils/skytilsmod/features/impl/misc/LockOrb.kt
```diff
@@ -37,30 +39,38 @@ class LockOrb {
...
-        ),
-        OVERFLUX("§5Overflux", 18.0, "OVERFLUX_POWER_ORB"), PLASMAFLUX("§d§lPlasmaflux", 20.0, "PLASMAFLUX_POWER_ORB");
+        RADIANT("§aRadiant", 18.0, "RADIANT_POWER_ORB"),
+        MANAFLUX("§9Mana Flux", 18.0, "MANA_FLUX_POWER_ORB"),
+        OVERFLUX("§5Overflux", 18.0, "OVERFLUX_POWER_ORB"),
+        PLASMAFLUX("§d§lPlasmaflux", 20.0, "PLASMAFLUX_POWER_ORB");
+
+        fun getDurationRemaining(entity: EntityArmorStand): Int? {
+            val durationText = entity.displayName.formattedText.substring(
+                orbName.length + 1,
```
My-Name-Is-Jeff, line 66: might break in the future but ok

My-Name-Is-Jeff (changes requested):

src/main/kotlin/skytils/skytilsmod/features/impl/misc/LockOrb.kt
lineargraph: If it does, at least it shouldn't crash

src/main/kotlin/skytils/skytilsmod/features/impl/misc/LockOrb.kt
lineargraph: Would be a change in behaviour. rn if it can't find a duration it just ignores the duration settings and blocks placing a new orb anyway. can be changed if you want it that way, should only matter if hypickel changes their format

src/main/kotlin/skytils/skytilsmod/core/Config.kt
lineargraph: alright, should i just preemptively remove the max, expecting future updates with even more powerful balls / pet items, or upgrade to 120 secs

src/main/kotlin/skytils/skytilsmod/features/impl/misc/LockOrb.kt
My-Name-Is-Jeff: technically it could be out of bounds but ok

src/main/kotlin/skytils/skytilsmod/core/Config.kt
My-Name-Is-Jeff: I think remove the max but idk if vigilance will work with that

lineargraph: It does not work with just one limit. Ig ill just put 120 as limit for now

My-Name-Is-Jeff (approved):

src/main/kotlin/skytils/skytilsmod/core/Config.kt
Sychic: Default vigilance max is 100 and it'll be that if you don't specify one

lineargraph: Actually it was 0 in my tests (which resulted in a crash cause the range 1..0 isnt valid)

My-Name-Is-Jeff: I think 100 is the default max for PERCENTAGE_SLIDER

PR #128, opened 2022-01-09
new style and more accurate total secrets count

BongleXD:
change the style and make the total secrets count more accurate


BongleXD:
shows like this

PR #137, opened 2022-01-12
more accurate speed score and more hiders

BongleXD:


Sychic:
image
Still at bit wrong at 186 overtime. Exact overtime was 185.584

BongleXD:
> image Still at bit wrong at 186 overtime. Exact overtime was 185.584

so i said more accurate there'll be 1 score gap sometime

BongleXD:
it should be more accurate at other floor like f5 f7 master modes

PR #158, opened 2022-02-17
Add King Waypoint

walkerselby:
The King is an equally important area to have as a waypoint and marked on the map. Unfortunately, Skyblock does not have a custom location for it like other areas in the Crystal Hollows.

To counteract this, besides setting it manually, I have added a check for messaging starting with "[NPC] King Yolkar" to set the location (and a config to toggle this behavior, if you only want to add King Yolkar manually).


My-Name-Is-Jeff:
We should really redo the waypoint code but for now this is fine

PR #161, opened 2022-02-22
Attempt to fix Dungeons Speed Score

zealerker22:
The speed score calculation algorithm is now in sync with the wiki one, along with changes for dungeon minimum time requirements.
However, this new algorithm is only tested with 50+ F1~F6 runs (not sure if the same algorithm applies for F7 or above) but this should fix 99+% of incorrect speed scores when the run goes over 10 mins.


My-Name-Is-Jeff:
I think we tested it and f7 was 12 min

zealerker22:
> I think we tested it and f7 was 12 min

that means 12 min 13 s for F7 should be 99 instead of 100? cool

My-Name-Is-Jeff:
Maybe they changed it recently

zealerker22:
> I think we tested it and f7 was 12 min
What about the Jan 6th edit in https://hypixel.net/threads/december-17-some-dungeon-qol.4715260/?
Was that reverted?

My-Name-Is-Jeff:
We tested it like a week after and it was 12 min lmao

src/main/kotlin/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
```diff
@@ -179,12 +179,13 @@ object ScoreCalculation {
...
         when {
-            time <= 480.0 -> 100.0
-            time <= 580.0 -> 148 - 0.1 * time
-            time <= 980 -> 119 - 0.05 * time
-            time < 3060 -> 102 - 1.0 / 30.0 * time
+            time < 492.0 -> 100.0
+            time < 600.0 -> 140 - time / 12
+            time < 840.0 -> 115 - time / 24
+            time < 1140.0 -> 108 - time / 30
+            time < 3940.0 -> 98.5 - 1.0 / 40.0 * time
```
Sychic: This should just be `98.5 - time / 40`

Sychic (changes requested):
One weird formatting thing but otherwise lgtm

src/main/kotlin/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
zealerker22: Lines 183-185 are meant to be integer divisions while line 186 does care about floating-point precisions:
if time is actually a floating number, then change scores in lines 183-185 from "time / [number]" to "time.toInt() / [number]" or whatever that forces integer division(along with your suggestion).
The reason behind integer divisions is Skyblock Dungeons stripping the remainder of speed score timestamps rather than rounding up/down accordingly(For example, a 10m 13s run, a 10m 18s run, and a 10m 22s run in F3 all give 99 speed pts even if speed score timestamps are actually 100/99/98 for 10m 0s/10m 12s/10m 24s[lmk if that's no longer the case anymore]).
Feel free to make changes as you see fit ig(I'm new to Kotlin after all).

PR #162, opened 2022-02-23
Update Config.kt

TheeUnwanted:
uwu rawr rawr


Sychic (approved):
Might want to make it red

shuuyu:
Make it red because they said uwu

PR #169, opened 2022-03-07
Please work(Party Finder Stats)

TakoTheDev:
If this doesnt work, i am giving up

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
+++ b/src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
@@ -18,21 +18,36 @@
 package skytils.skytilsmod.listeners
 
 import gg.essential.universal.UChat
+import gg.essential.universal.wrappers.message.UMessage
+import gg.essential.universal.wrappers.message.UTextComponent
+import net.minecraft.event.ClickEvent
 import net.minecraft.util.ChatComponentText
 import net.minecraft.util.EnumChatFormatting
 import net.minecraftforge.client.ClientCommandHandler
 import net.minecraftforge.client.event.ClientChatReceivedEvent
 import net.minecraftforge.fml.common.eventhandler.EventPriority
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
+import skytils.hylin.extension.getString
+import skytils.hylin.mojang.AshconException
+import skytils.hylin.request.HypixelAPIException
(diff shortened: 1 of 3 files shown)


src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
```diff
@@ -46,6 +61,95 @@ class ChatListener {
...
+                    }
+                    return
+                }
+            }
+            val joinedd = dungeon_finder_join_pattern.matcher(unformatted)
+            if (joinedd.find()) {
+                val username = joinedd.group(1)
+                if(username.equals(mc.thePlayer.name)) {
+                    try {
+                        Thread.sleep(1000)
```
Sychic, line 114: This is on the main thread, you should not be calling thread.sleep here.

src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
```diff
@@ -211,6 +315,262 @@ class ChatListener {
...
+                        } else if (itemName.contains("Axe of the Shredded")) {
+                            hasAots = true
+                        } else if (itemName.contains("Giant's Sword")) {
+                            hasGiantSword = true
+                        } else if (itemName.contains("Shadow Fury")) {
+                            hasShadowFury = true
+                        } else if (itemName.contains("Flower of Truth")) {
+                            hasFlowerOfTruth = true
+                        }
+                    }
```
Sychic, line 451: This should be a when block

src/main/kotlin/skytils/skytilsmod/listeners/ChatListener.kt
```diff
@@ -211,6 +315,262 @@ class ChatListener {
...
+                            append("§2Giant's Sword: §a§l●\n")
+                        } else if(hasAots) {
+                            append("§aAots: §a§l●\n")
+                        } else if(hasShadowFury) {
+                            append("§8Shadow Fury: §a§l●\n")
+                        } else if(hasFlowerOfTruth) {
+                            append("§cFot: §a§l●\n")
+                        } else {
+                            append("§2Giant's Sword: §c§l●\n")
+                        }
```
Sychic, line 487: This should be a when block

TakoTheDev:
Doesnt work since i could finally test it after downgrading to pre2

Sychic (changes requested):

PR #185, opened 2022-03-24
Feature/container sell value

FluxCapacitor2:
I added an overlay that renders when a chest, ender chest page, backpack, or minion is open and shows the container's 20 most valuable items along with the combined value of all items in the container. It can be moved and scaled just like a regular `GuiElement`, but it is rendered after the container background is drawn to prevent the text from being dimmed (similar to the `ChestProfit` element). When the overlay is rendered over a minion GUI, minion upgrades and fuels aren't included in the value.

* Should I add config options to control the maximum amount of items to display?
* Should I include things like enchantments and HPBs in item prices?
* Should enchanted books be grouped by enchantments or by item name? Currently all items are grouped by item name, and enchanted books are only displayed as "Enchanted Book" instead of something like "Soul Eater I"
* Let me know if you want any formatting changes

Here's an example of this overlay being used while inside a minion GUI:
image

Example of the truncation:
image

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
+++ b/src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
@@ -154,7 +154,7 @@ class ContainerSellValue {
             }
         }
 
-        if(slots.isEmpty()) return
+        if(slots.isEmpty() || distinctItems.isEmpty() || distinctItems.all { it.value.lowestBIN == 0.0 }) return
 
         // Sort the items from most to least valuable and convert them into a readable format
         val textLines = distinctItems.entries.asSequence()
--- a/src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
+++ b/src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
@@ -18,9 +18,10 @@
 
 package skytils.skytilsmod.features.impl.misc
 
+import gg.essential.universal.UGraphics
+import gg.essential.universal.UMatrixStack
 import gg.essential.universal.UResolution
 import net.minecraft.client.gui.inventory.GuiChest
-import net.minecraft.client.renderer.GlStateManager
 import net.minecraft.inventory.ContainerChest
 import net.minecraft.item.ItemStack
 import net.minecraftforge.client.event.GuiScreenEvent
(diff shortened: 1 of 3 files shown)


src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
```diff
@@ -0,0 +1,193 @@
...
+import skytils.skytilsmod.utils.graphics.SmartFontRenderer
+import skytils.skytilsmod.utils.graphics.colors.CommonColors
+import kotlin.math.roundToInt
+
+/**
+ * Provides the functionality for an overlay that is rendered on top of chests, minions, ender chest pages,
+ * and backpacks that shows the top items in the container by value and the container's total value.
+ * @author FluxCapacitor2
+ */
+class ContainerSellValue {
```
Sychic: Since this is a singleton, it should be an object

src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
```diff
@@ -0,0 +1,193 @@
...
+
+        init {
+            Skytils.guiManager.registerElement(this)
+        }
+    }
+
+    companion object {
+        val element = SellValueDisplay()
+        // The max amount of BIN prices to show in the list. If there are more items, the list will be truncated to this size.
+        const val MAX_DISPLAYED_ITEMS = 20
```
Sychic: maybe this should be a setting

src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
```diff
@@ -0,0 +1,193 @@
...
+
+        textLines.forEachIndexed { i, str ->
+            ScreenRenderer.fontRenderer.drawString(
+                str,
+                element.textPosX, (i * ScreenRenderer.fontRenderer.FONT_HEIGHT).toFloat(),
+                CommonColors.WHITE, element.alignment, SmartFontRenderer.TextShadow.NORMAL
+            )
+        }
+
+        GlStateManager.popMatrix()
```
Sychic: Ideally this should be using UMatrixStack

Sychic:
Enchants would definitely make more sense as the enchant rather than just `Enchanted Book`

src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
Sychic:
```suggestion
object ContainerSellValue {
```

FluxCapacitor2:
Fixed, enchants now look like this:
image
Enchanted books with only one enchant use the first line of their lore to determine which string to display, and if a book has multiple enchants, it is still labeled as "Enchanted Book"

PR #187, opened 2022-03-25
changed/added smaller features

TakoTheDev:
Config:
- Made Party Finder stats inside the Dungeons category(inside the code)
- added 270 and 300 title message config
- changed 270 and 300 to be not inside of each other(so it's 270, 270, 300, 300 instead of 270, 300, 270, 300)
- added change all to same color modes(Normal, Left/Right, Monkey) *can be renamed*
- added dependencies for 270 and 300 stuff and change to same color

Party Finder Stats:
- Floored cata for cleaner message
- added default message when no items
- made it ignore Entrance(cause no one cares)
- made all \n be on the right side
- Removed Misc message part (To make it cleaner)
- removed not needed {}

Score Calculation:
- custom 270 and 300 title message
- shortened the 270 and 300 chat message(no prefix)
- make mimic killed message look cleaner(Mimic Killed! instead of that abomination)

Change All to same color:
- added 3 modes:
- Normal has positive and negative numbers(currently all white)
- left/right has left/right click(currently all white)
- Monkey has Left click only numbers(1 click green:2,3 clicks yellow;4 clicks red)

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -18,8 +18,8 @@
 package skytils.skytilsmod.core
 
 import gg.essential.api.EssentialAPI
-import gg.essential.universal.UDesktop
 import gg.essential.elementa.utils.withAlpha
+import gg.essential.universal.UDesktop
 import gg.essential.vigilance.Vigilant
 import gg.essential.vigilance.data.Category
 import gg.essential.vigilance.data.Property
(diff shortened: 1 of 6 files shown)


TakoTheDev:
Removing the client import may have caused an issue, since the ?bot? failed to compile it

TakoTheDev:
forgot to upload changes will be done tomorrow morning?

Sychic:
Not really useful, most of these can be created using existing features (ie a lot of them can be notifications)

PR #196, opened 2022-04-01
Container sell value improvements

FluxCapacitor2:
New features:
- Include item modifiers in prices: enchantments, hot potato books, recombobulators, and master stars (suggested by azured)
- Add bazaar prices to `AuctionData` and use them to give more accurate item prices (also suggested by azured)
- Add compatibility with NEU's storage overlay (suggestion #1956)
- Prices are only calculated ~4 times per second and cached in a list for a very tiny performance improvement

Other things to note:
- All changes have been tested with and without NEU storage overlay. The only oversight I can think of is, dungeon items with random stats can have their value increased by about 15M because they commonly have scavenger 5 on them.
- Clearing and adding new items to the list may cause a CME but I haven't encountered that yet. If that is a concern, what is the best practice for mitigating that? Should I change the type of `MutableList` or should I just wrap the rendering block with a `runCatching`?


src/main/kotlin/skytils/skytilsmod/features/impl/handlers/AuctionData.kt
```diff
@@ -81,6 +86,13 @@ class AuctionData {
                 }
             }
         }
+        fixedRateTimer(name = "Skytils-FetchBazaarData", period = 10 * 60 * 1000L) {
+            if(Skytils.config.fetchLowestBINPrices && Skytils.config.containerSellValue) {
+                Skytils.hylinAPI.getBazaarData().whenComplete {
```
My-Name-Is-Jeff: idk if this is gonna break hypixel but hopefully not? i can just put all the bz prices on the lowestbin cause that's what it already does for certain whitelisted items

src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
```diff
@@ -109,18 +110,45 @@ object ContainerSellValue {
         }
     }
 
-    private fun shouldRenderOverlay(chestName: String): Boolean {
+    private fun isChestNameValid(chestName: String): Boolean {
         return (!inDungeons && (chestName == "Chest" || chestName == "Large Chest"))
                 || chestName.contains("Backpack")
                 || chestName.startsWith("Ender Chest (")
                 || (chestName.contains("Minion") && SBInfo.mode == SkyblockIsland.PrivateIsland.mode)
```
My-Name-Is-Jeff: idk how I didn't catch this before but I think this would trigger in the crafting recipe for minions

src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
```diff
@@ -135,21 +163,86 @@ object ContainerSellValue {
...
+     * Renders the sell value overlay.
+     * @see onPostBackgroundDrawn
+     * @see onGuiScreenDrawPre
+     */
+    fun renderGuiComponent() {
+        // Translate and scale manually because we're not rendering inside the GuiElement#render() method
+        val stack = UMatrixStack()
+        stack.push()
+        stack.translate(element.actualX, element.actualY, 0f)
+        stack.scale(element.scale, element.scale, 0f)
```
My-Name-Is-Jeff, line 214: i think this will result in the same bug as the one for ch map? not sure

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/AuctionData.kt
Sychic: Ideally we should try to reduce our api queries when possible

src/main/kotlin/skytils/skytilsmod/features/impl/misc/ContainerSellValue.kt
FluxCapacitor2: I haven't experienced any issues with it and it conforms to the size of the box in the location edit gui, so it should be fine.

PR #201, opened 2022-04-10
Added prefixes and more stuff

AzuredBlue:
- Added prefixes to most messages 
- Changed shield timer in /st locations 
- Some changes to party finder stats to make it more similar to /skytilscata 
- Added an actual command usage for /skytilscalcxp 
- Added some colors to /st ping because why not
- Changed some dungeon timer messages to make them match the color of the gui
- Removed real time from dungeon timer and added a portal time

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/Skytils.kt
+++ b/src/main/kotlin/skytils/skytilsmod/Skytils.kt
@@ -173,6 +173,10 @@ class Skytils {
         val areaRegex = Regex("§r§b§l(?<area>[\\w]+): §r§7(?<loc>[\\w ]+)§r")
 
         var domain = "api.skytils.gg"
+
+        var prefix = "§9§lSkytils §8»"
+        var successPrefix = "§a§lSkytils §8»"
+        var failPrefix = "§c§lSkytils §8»"
     }
 
     @Mod.EventHandler
--- a/src/main/kotlin/skytils/skytilsmod/Skytils.kt
+++ b/src/main/kotlin/skytils/skytilsmod/Skytils.kt
@@ -174,7 +174,7 @@ class Skytils {
 
         var domain = "api.skytils.gg"
 
-        val prefix = "§9§lSkytils §8»"
+        const val prefix = "§9§lSkytils §8»"
         val successPrefix = "§a§lSkytils §8»"
         val failPrefix = "§c§lSkytils §8»"
     }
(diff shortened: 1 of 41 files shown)


src/main/kotlin/skytils/skytilsmod/Skytils.kt
```diff
@@ -173,6 +173,10 @@ class Skytils {
         val areaRegex = Regex("§r§b§l(?<area>[\\w]+): §r§7(?<loc>[\\w ]+)§r")
 
         var domain = "api.skytils.gg"
+
```
My-Name-Is-Jeff, line 176: why is it var not val (const val maybe?)

My-Name-Is-Jeff: also idk if it should be like a function but it's fine

src/main/kotlin/skytils/skytilsmod/Skytils.kt
```diff
@@ -173,6 +173,10 @@ class Skytils {
...
+        val prefix = "§9§lSkytils §8»"
```
My-Name-Is-Jeff:
```suggestion
        const val prefix = "§9§lSkytils §8»"
```

src/main/kotlin/skytils/skytilsmod/Skytils.kt
```diff
@@ -173,6 +173,10 @@ class Skytils {
...
+        val successPrefix = "§a§lSkytils §8»"
```
My-Name-Is-Jeff:
```suggestion
        const val successPrefix = "§a§lSkytils §8»"
```

src/main/kotlin/skytils/skytilsmod/Skytils.kt
```diff
@@ -173,6 +173,10 @@ class Skytils {
...
+        val failPrefix = "§c§lSkytils §8»"
```
My-Name-Is-Jeff:
```suggestion
        const val failPrefix = "§c§lSkytils §8»"
```

My-Name-Is-Jeff (approved):

src/main/kotlin/skytils/skytilsmod/features/impl/dungeons/DungeonTimer.kt
```diff
@@ -227,6 +232,16 @@ class DungeonTimer {
                                 )
                             }s"
                         )
+                    if (bloodClearTime != -1L)
+                        if (!DungeonFeatures.dungeonFloor.equals("E"))
+                        add(
+                            "§dPortal: ${
+                                diff(
+                                    if (bossEntryTime == -1L) System.currentTimeMillis() else bossEntryTime,
```
My-Name-Is-Jeff: if they don't enter boss it'll keep going up even if they fail the run but ok

PR #202, opened 2022-04-11
Mining: Small Additions and Fixes

walkerselby:
Add Pristine Spam Hider
Add Pristine Spam Hider Config
Fix an issue with Goblin Queen's Den not automatically propagating on the map
Add Corleone Marker to Map

### DIFF
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -2626,6 +2626,14 @@ object Config : Vigilant(
     )
     var pristineHider = 0
 
+    @Property(
+        type = PropertyType.SELECTOR, name = "Wind Direction Hider",
+        description = "Removes Gone With the Wind direction change messages.",
+        category = "Spam", subcategory = "Miscellaneous",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
+    var windHider = 0
+
     init {
         addDependency("showEtherwarpTeleportPosColor", "showEtherwarpTeleportPos")
 
--- a/src/main/kotlin/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/skytils/skytilsmod/core/Config.kt
@@ -2618,6 +2618,14 @@ object Config : Vigilant(
     )
     var compactHider = 0
 
+    @Property(
+        type = PropertyType.SELECTOR, name = "Pristine Hider",
+        description = "Removes Pristine messages from mining.",
+        category = "Spam", subcategory = "Miscellaneous",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
+    var pristineHider = 0
+
     init {
         addDependency("showEtherwarpTeleportPosColor", "showEtherwarpTeleportPos")
 
--- a/src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
+++ b/src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
@@ -519,6 +519,13 @@ class SpamHider : PersistentSave(File(Skytils.modDir, "spamhider.json")) {
                     }
                 }
 
+                // Pristine
+                unformatted.startsWith("PRISTINE! You found") -> {
+                    when (Skytils.config.pristineHider) {
+                        1, 2 -> cancelChatPacket(event, Skytils.config.pristineHider == 2)
+                    }
+                }
+
                 // Blocks in the way
                 unformatted.contains("There are blocks in the way") -> {
                     when (Skytils.config.inTheWayHider) {
--- a/src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
+++ b/src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
@@ -339,11 +339,16 @@ class SpamHider : PersistentSave(File(Skytils.modDir, "spamhider.json")) {
                 }
 
                 // Mining Abilities
-                formatted.startsWith("§r§a§r§6") && formatted.endsWith("§r§ais now available!§r") -> {
+                (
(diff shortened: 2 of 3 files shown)


src/main/kotlin/skytils/skytilsmod/core/Config.kt
```diff
@@ -2618,6 +2618,22 @@ object Config : Vigilant(
     )
     var compactHider = 0
 
+    @Property(
+        type = PropertyType.SELECTOR, name = "Pristine Hider",
+        description = "Removes Pristine messages from mining.",
+        category = "Spam", subcategory = "Miscellaneous",
+        options = ["Normal", "Hidden", "Separate GUI"]
+    )
+    var pristineHider = 0
```
My-Name-Is-Jeff, line 2627: I think there already is a pristine hider I'm not too sure

walkerselby: I looked through ingame and through the code and could not find it. I might have missed it, though.

src/main/kotlin/skytils/skytilsmod/features/impl/mining/MiningFeatures.kt
```diff
@@ -369,11 +369,12 @@ class MiningFeatures {
         enum class Locations(val displayName: String, val id: String, val color: Int, val size: Int = 50) {
             LostPrecursorCity("§fLost Precursor City", "internal_city", ColorFactory.WHITE.rgb),
             JungleTemple("§aJungle Temple", "internal_temple", ColorFactory.GREEN.rgb),
-            GoblinQueensDen("§eGoblin Queens Den", "internal_den", ColorFactory.YELLOW.rgb),
+            GoblinQueensDen("§eGoblin Queen's Den", "internal_den", ColorFactory.YELLOW.rgb),
```
My-Name-Is-Jeff, line 372: I think this still requires a change to https://github.com/Skytils/SkytilsMod/blob/e098fcc50a6f2999e372e7c39c7213c2aed320cf/src/main/kotlin/skytils/skytilsmod/utils/SBInfo.kt#L65

src/main/kotlin/skytils/skytilsmod/core/Config.kt
Sychic: https://github.com/Skytils/SkytilsMod-Data/blob/main/SpamFilters.json#L14-L19

walkerselby: @Sychic Gotcha. Is there a reason why some stuff, like that, is within the Spam Filters system from the Skytils-Data repo and the Spam Filters menu ingame, rather than the general spam config section? 

From a user standpoint, seeing compact be in the spam section and not pristine was confusing, leading me to think that it was not an option. I generally thought of the spam filters for custom things, rather than gameplay features, though that, in your team's mind, might not be the intended usecase.

Sychic: The main reason for creating the repo filters was so then we could roll out new spam hiders remotely without having to release a new version of the mod. Though the custom spam hiders has mostly overtaken this usage since anyone can create any spam hider they want.

src/main/kotlin/skytils/skytilsmod/features/impl/mining/MiningFeatures.kt
walkerselby: I think it is fine? Unless you see something that I am missing. It works ingame and using a Regex checker, it detects "Goblin Queen's Den" fine.

src/main/kotlin/skytils/skytilsmod/core/Config.kt
walkerselby: Gotcha. I am fine not having it added, if that is your guy's preference (and I can remove it from this PR), but I would also think that it would fit with compact in the actual menu. 

I do see the desire for adding things without a new version of the mod, I maybe would add a link to the spam filters in the spam section then, as it was not apparent to me that it would be in that section, as well as talking to a few friends. Just seems like some features will get missed by users. 

In my opinion, it is worth having most of the general spam features in the actual config, even if that requires a new mod version, and having the spam filters section mostly be custom ones, especially since Skytils has such a great updating system. But of course, that is just my opinion, feel free to do whatever you think is best for the vision of the mod.

walkerselby: Looking again -- I do see "Legacy Hiders" inside the Spam Filters menu. Maybe at some point Spam migrates entirely to the Spam Filter system, including allowing custom, so there is not options duplicated in two menus.

Sychic: > Looking again -- I do see "Legacy Hiders" inside the Spam Filters menu. Maybe at some point Spam migrates entirely to the Spam Filter system, including allowing custom, so there is not options duplicated in two menus.

That would be the goal but waiting until we have search for it, the only real reason for a legacy hider now is for toasts (as there's no good way to custom add it)

src/main/kotlin/skytils/skytilsmod/features/impl/handlers/SpamHider.kt
```diff
@@ -339,12 +339,17 @@ class SpamHider : PersistentSave(File(Skytils.modDir, "spamhider.json")) {
                 }
 
                 // Mining Abilities
-                formatted.startsWith("§r§a§r§6") && formatted.endsWith("§r§ais now available!§r") -> {
+                (
+                        (formatted.startsWith("§r§a§r§6") && formatted.endsWith("§r§ais now available!§r"))
+                                || formatted.contains("§r§cYour Mining Speed Boost has expired!§r")
+                                || formatted.contains("§r§aYou used your §r§6Mining Speed Boost §r§aPickaxe Ability!§r")
+                        ) -> {
```
Sychic, line 346: Perhaps this should be a Regex?

Sychic (approved):

PR #216, opened 2022-04-24
Container Sell Value: Fix master star value

FluxCapacitor2:
Because of the changes to the way master stars are displayed in item names, the item value calculation broke.

I've also excluded item modifiers from randomized dungeon items (logic stolen from the "highlight salvageable items" feature), primarily because of Scavenger 5 being worth more than the item itself.


src/main/kotlin/skytils/skytilsmod/utils/ItemUtil.kt
```diff
@@ -229,4 +230,14 @@ object ItemUtil {
 
     fun getStarCount(extraAttributes: NBTTagCompound) =
         extraAttributes.getInteger("upgrade_level") + extraAttributes.getInteger("dungeon_item_level")
+
+    fun isSalvageable(stack: ItemStack): Boolean {
+        val extraAttr = getExtraAttributes(stack)
+        val sbId = getSkyBlockItemID(extraAttr)
+        return sbId != "ICE_SPRAY_WAND" && extraAttr != null && getStarCount(extraAttr) == 0
+                && !extraAttr.hasKey("dungeon_item") && (extraAttr.hasKey("baseStatBoostPercentage") || getItemLore(
```
My-Name-Is-Jeff: remove extra lore check as they added the nbt tag back, probably move the baseStatBoosPercentage up to the front

My-Name-Is-Jeff (changes requested):

PR #230, opened 2022-05-20
Update DSM/SBE Regex

walkerselby:
Make the first group non-capturing, as it is not used, to save performance slightly.

Fixed an issue with regex where it processes as a negative number when it will never be a negative number. "@-" is always included, but does not signify a negative coordinate. This prevents existing issues where the location is placed off the map. 

- whalker


My-Name-Is-Jeff:
Thanks!

PR #234, opened 2022-06-05
Low health vignette

FluxCapacitor2:
Renders a vignette, just like when the player is outside of the world border, when the player's health is low. It gets more opaque as the player's health decreases. Does not conflict with the world border fix ("Red screen fix").

Resolves suggestion 2347.


src/main/kotlin/gg/skytils/skytilsmod/utils/RenderUtil.kt
```diff
@@ -758,6 +758,43 @@ object RenderUtil {
...
+        UGraphics.color4f(
+            (1f - (color.red / 255f)) * (color.alpha / 255f),
+            (1f - (color.green / 255f)) * (color.alpha / 255f),
+            (1f - (color.blue / 255f)) * (color.alpha / 255f),
+            1f
+        )
+
+        val sr = UResolution
+
+        UGraphics.bindTexture(0, vignetteTexPath)
```
FluxCapacitor2, line 787: To be honest I have no idea what the first parameter (`index`) actually does, but setting it to 0 has not caused any problems

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/MiscFeatures.kt
```diff
@@ -149,7 +150,7 @@ class MiscFeatures {
                 }
             }
             if (unformatted == "Your zapper is temporarily fatigued!") {
-                val duration = Duration.milliseconds(blockZapperCooldownExpiration - System.currentTimeMillis())
+                val duration = (blockZapperCooldownExpiration - System.currentTimeMillis()).milliseconds
```
My-Name-Is-Jeff: This won't work cause essential

src/main/kotlin/gg/skytils/skytilsmod/utils/RenderUtil.kt
My-Name-Is-Jeff: I don't know if this method actually exists or not

My-Name-Is-Jeff (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/MiscFeatures.kt
FluxCapacitor2: IntelliJ auto corrected it without telling me :smile:

PR #236, opened 2022-06-10
fix some linux related issues

Tricked-dev:


src/main/java/SkytilsInstallerFrame.java
```diff
@@ -189,11 +189,22 @@ public void actionPerformed(ActionEvent e) {
 
     @Override
     public void mouseClicked(MouseEvent e) {
+        String forgeUrl = "http://files.minecraftforge.net/maven/net/minecraftforge/forge/index_1.8.9.html";
         if (e.getSource() == descriptionPanel.forge) {
             try {
-                Desktop.getDesktop().browse(new URI("http://files.minecraftforge.net/maven/net/minecraftforge/forge/index_1.8.9.html"));
-            } catch (IOException | URISyntaxException ex) {
+                Desktop.getDesktop().browse(new URI(forgeUrl));
```
Sychic, line 195: We should be fine to use UDesktop.browse instead

src/main/java/SkytilsInstallerFrame.java
```diff
@@ -84,7 +84,7 @@ public SkytilsInstallerFrame() {
             testPanel.add(chooser, BorderLayout.EAST);
 
             // Footer
-            footer = new FooterPanel(w / 2, h / 11, 16, 16, new Color(0x8193ff));
+            footer = new FooterPanel(w / 2 + (getOperatingSystem( ) == OperatingSystem.LINUX ?  60  : 0) , h / 11, 16, 16, new Color(0x8193ff));
```
Sychic, line 87: This seems like a hack that shouldn't be necessary, are you sure this is the case for the majority of linux users/fonts?

Sychic (changes requested):

src/main/java/SkytilsInstallerFrame.java
My-Name-Is-Jeff: No we should not

src/main/java/SkytilsInstallerFrame.java
Tricked-dev: its a hack-ish i dont think its possible to change the size based on font width and this way its almost guaranteed that everything is visible unlike before where you can't even see half of the cancel button

PR #237, opened 2022-06-11
Fix waypoint text rendering when the player is nearby

FluxCapacitor2:
When the player is nearby, waypoints' beacon beams aren't rendered, which means that `enableTexture2D()` is never called.


My-Name-Is-Jeff:
Put this in draw name tag I think

PR #258, opened 2022-09-04
Added "Bat Swarm Hider"

MrGcGamer:
I was unable to get the dev env working + I never used Kotlin before, but I think I did everything right non the less (hopefully).
(Not sure though, if I should have used `formatted.contains` instead of `unformatted.contains` ...it would be greatly appreciated, if someone could explain to which one is used in which case)


My-Name-Is-Jeff:
too minor, can be done with custom spam hider

PR #267, opened 2022-09-29
Fix crash in ContainerSellValue

lineargraph:


lineargraph:
Hi @My-Name-Is-Jeff any updates on this?

My-Name-Is-Jeff:
Can't repro

lineargraph:
I mean i could, and so could like everyone in our support channels i gave this to. And either way this isnt exactly like it would be worse to do this way either.

lineargraph:
It's nicer code in it's own right, it fixes a bug, and you *already* have a hacky solution that you needed to do because of NEUs storage gui, so why not try to be fully compatible with NEU.

PR #269, opened 2022-10-05
Fix CH waypoints shifts a lot when teleporting

DespairScent:
A little hacky, but still a working way


DespairScent:
Sometimes when lagging, the location in the scoreboard changes for more than 1 sec. after teleporting, so this needs improvement or a different approach.

DespairScent:
I found that location string updates every n20+2..5 world tick. The ticks seem to be in sync with the server, so this can be used.

Sychic:
This is an inaccurate way since it can still be de-synced via lag. This ties into another flaw with the system where it tries to update the location despite it already being set. The entire system should be rewritten to only set the waypoint once.

DespairScent:
> This is an inaccurate way since it can still be de-synced via lag. This ties into another flaw with the system where it tries to update the location despite it already being set. The entire system should be rewritten to only set the waypoint once.

About setting waypoint once. There is still a problem when you are not really there yet, but in the scoreboard it shows the location. This is still useful for finding actual location position, but it will require manually setting waypoint later. For yourself, or for sharing.

PR #275, opened 2022-10-25
Fix inviting multiple players in one message

CuzImClicks:
Breefing, Ealman and DeathStreeks complained about it in their streams a lot
(When inviting multiple players in one command hypixel creates an invalid invite)

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/commands/impl/PartyCommand.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/commands/impl/PartyCommand.kt
@@ -30,12 +30,12 @@ object PartyCommand : BaseCommand("p", listOf("party")) {
             return
         }
         if (!args.first().endsWith(",") && !((args.first() == "invite" || args.first() == "i") && (args[1].endsWith(",") || args.size > 2))) {
-            UChat.chat("/p ${args.joinToString(" ")}")
+            UChat.say("/p ${args.joinToString(" ")}")
             return
         } else {
             Thread {
                 args.filter { it != "invite" && it != "i" }.forEach {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/gg/skytils/skytilsmod/commands/impl/PartyCommand.kt
```diff
@@ -0,0 +1,44 @@
...
+        if (!args.first().endsWith(",")) {
+            UChat.say("/p ${args.joinToString(" ")}")
+            return
+        } else {
+            Thread {
+                args.forEach {
+                    UChat.say("/p ${it.replace(",", "").replace(" ", "")}")
+                    Thread.sleep(400)
+                }
+            }.start()
```
TakoTheDev: Skytils.sendMessageQueue.add("/command to run") exists

TakoTheDev: Also why replace " "?

TakoTheDev (commented):
Just make a list of sub-commands (listOf("i", "invite"))

src/main/kotlin/gg/skytils/skytilsmod/commands/impl/PartyCommand.kt
CuzImClicks: > Skytils.sendMessageQueue.add("/command to run") exists

uses a delay of 250ms while hypixel does not allow inviting players that quickly

> Also why replace " "?

safety

My-Name-Is-Jeff:
closed on accident but i don't like how it replaces any commands

PR #298, opened 2022-12-03
Prevent NumberFormatException crash in Waypoint GUI

appable0:
Prevents a NumberFormatException crash in Waypoint GUI by casting non-numeric coordinates (e.g. "-" or "1-") to 0. 

Also attempts to notify user of formatting issue by marking coordinate text input fields in red when an invalid input is entered (and resetting when the invalid input is corrected).

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/utils/ElementaUtil.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/utils/ElementaUtil.kt
@@ -23,6 +23,7 @@ import gg.essential.elementa.components.UIContainer
 import gg.essential.elementa.components.input.UITextInput
 import gg.essential.universal.UKeyboard
 import gg.essential.vigilance.gui.settings.SettingComponent
+import java.awt.Color
 
 /**
  * When the tab key is pressed in this text field, [other] will grab the window's focus.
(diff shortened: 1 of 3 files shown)


src/main/kotlin/gg/skytils/skytilsmod/utils/ElementaUtil.kt
```diff
@@ -42,6 +43,19 @@ fun UITextInput.limitToNumericalCharacters() = apply {
...
+ */
+fun UITextInput.colorIfNumeric(validColor: Color, invalidColor: Color) = apply {
+    onKeyType { _, _ ->
+        if (getText().isInteger()) setColor(validColor)
+    }
+    onFocusLost {
+        if (!getText().isInteger()) setColor(invalidColor)
+    }
+}
+
```
Sychic, line 53: Why not make it also do invalid as you're typing?

appable0: I thought it might be annoying if you're trying to type "-x" and it briefly shows up as red—maybe that's alright?

Sychic: I mean it would make sense when it's only the `-` sign.

appable0: Fair point, updated to show invalid inputs immediately (also attempted to squash but may have failed, unsure).

PR #308, opened 2022-12-30
Added Free Slots Display

TakoTheDev:
Suggestion: id 2336
kat = cool
Also wasn't sure if this should be in ItemFeatures.kt or MiscFeatures.kt
Looks like this:
image


My-Name-Is-Jeff:
Azured says this is useless

PR #311, opened 2023-01-03
Adding comment for decompilers about the frag bot verify logic

hannibal002:
No more confusion because we have now for the third time the same comment that it is safe and not a scam.


My-Name-Is-Jeff:
fr

PR #314, opened 2023-01-06
Added toggle to floor completions and armour in party finder stats

Ownwn:


Sychic:
Like with tako, can we try to avoid mixing too many features in a single pr?

PR #317, opened 2023-01-31
Added MC version number to coremod AsmHelperLoadingPlugin

hannibal002:
Fixing this error message:
`The coremod gg.skytils.skytilsmod.tweaker.SkytilsLoadingPlugin does not have a MCVersion annotation, it may cause issues with this version of Minecraft `


My-Name-Is-Jeff:
There was a reason that I forgot about why this line was removed

PR #327, opened 2023-02-15
Added Sam Scythe/ Garden Scythe highlighting

CarsCupcake:
This will add an option to highlight blocks which will be brocken by sam scythe/garden scythe


TakoTheDev:
You should probably press ctrl + shift or alt + L (I hope that was the shortcut, you can probably also right-click and find it in that menu), as that automatically formats the code correctly, cause I'm seeing some missing spaces

Issue #335, opened 2023-02-28
Party manager


My-Name-Is-Jeff:
https://discord.com/channels/807302538558308352/949545945106878504/1044647537971318794

My-Name-Is-Jeff:
Closed by #343

PR #343, opened 2023-03-04
The Breefing Thing

TakoTheDev:
For this weird party manager thing in #335 ig (code probably sucks idk)
Also some screenshots (outdated, but really similar to current one):
image
image
image


src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+    private val leaderPattern = Regex("^§eParty Leader: (?<rank>.+? |§7)?(?<name>\\w+) §r(?<status>§a|§c)●§r")
+    private val playerPattern = Regex("(?<rank>.+? |§7)?(?<name>\\w+)§r(?<status>§a|§c) ● ")
+    private val party = mutableListOf<PartyMember>()
+
+    //0 = not awaiting, 1 = awaiting 2nd delimiter, 2 = awaiting 1st delimiter
+    private var awaitingMessage = 0
+
+    @SubscribeEvent
+    fun onCommandRun(event: SendChatMessageEvent) {
+        if (!Utils.inSkyblock || !Skytils.config.partyAddons) return
```
My-Name-Is-Jeff: maybe do on hypixel?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+        if (Utils.equalsOneOf(event.message, "/pl", "/party list", "/p list", "/party l")) {
```
My-Name-Is-Jeff: are all of these hypixel commands?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+
+    @SubscribeEvent
+    fun onChat(event: ClientChatReceivedEvent) {
+        if (!Utils.inSkyblock || event.type == 2.toByte() || !Skytils.config.partyAddons) return
+        val message = event.message.formattedText
+
+        if (message == "§f§r" && awaitingMessage != 0) {
+            event.isCanceled = true
+        } else if (message.startsWith("§6Party Members")) {
+            val partyStart = partyStartPattern.find(message)
```
My-Name-Is-Jeff: if you don't care about the groups do like containsMatchIn or matches

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+                val name = leader.groups["name"]?.value ?: return
+                val status = leader.groups["status"]?.value ?: return
+                val rank = leader.groups["rank"]?.value ?: return
+
+                party.add(PartyMember(name, PartyMemberType.LEADER, status, rank))
+                event.isCanceled = true
+            }
+        } else if (message.startsWith("§eParty Moderators: ")) {
+            playerPattern.findAll(message.substringAfter("§eParty Moderators: ")).forEach {
+                val name = it.groups["name"]?.value ?: return
```
My-Name-Is-Jeff: not possible for these to be null, assert non-null instead

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+            event.isCanceled = true
+        } else if (message.startsWith("§cYou are not currently in a party.") && awaitingMessage != 0) {
+            party.clear()
+        } else if (message == "§9§m-----------------------------------------------------§r" && awaitingMessage != 0) {
+            awaitingMessage--
+            if (awaitingMessage == 1 || party.isEmpty()) return
+
+            val component = UMessage("§aParty members (${party.size})\n")
+
+            val player = party.first { it.name == Skytils.mc.thePlayer.name }
```
My-Name-Is-Jeff: codestyle: 
static import Skytils.mc
pick a different name that's not player (maybe like "me", "self" or something idk)

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+                    )
+                )
+            }
+
+            val partyLeader = party.first { it.type == PartyMemberType.LEADER }
+            component.append(
+                "\n${partyLeader.status}➡§r ${partyLeader.rank}${partyLeader.name}"
+            )
+
+            if (party.any { it.type == PartyMemberType.MODERATOR }) component.append("\n§eMods")
```
My-Name-Is-Jeff: iterates through the list twice for no reason, maybe save the result of the filter

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+                ).append(
+                    createButton(
+                        "§4[☠]",
+                        "/p kick ${it.name}",
+                        "§4Kick ${it.name}"
+                    )
+                )
+            }
+
+            if (party.any { it.type == PartyMemberType.MEMBER }) component.append("\n§eMembers")
```
My-Name-Is-Jeff: iterates through the list twice for no reason, maybe save the result of the filter

My-Name-Is-Jeff (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
TakoTheDev: yea all /pl variations, not sure if I still even need it as I initially used it for something else but idc

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
TakoTheDev: yea true am kinda dumb ngl

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
TakoTheDev: val mods = party.filter(...)
if (mods.isNotEmpty()) "Mods"
assuming something like this right?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+        if (message == "§f§r" && awaitingMessage != 0) {
+            event.isCanceled = true
+        } else if (partyStartPattern.matches(message)) {
+            party.clear()
+            event.isCanceled = true
+        } else if (message.startsWith("§eParty Leader")) {
+            val leader = leaderPattern.find(message)
+
+            if (leader != null) {
+                val name = leader.groups["name"]?.value!!
```
My-Name-Is-Jeff: bro why

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+                        PartyMemberType.MODERATOR,
+                        status,
+                        rank
+                    )
+                )
+            }
+            event.isCanceled = true
+        } else if (message.startsWith("§eParty Members: ")) {
+            playerPattern.findAll(message.substringAfter("§eParty Members: ")).forEach {
+                val name = it.groups["name"]?.value ?: return
```
My-Name-Is-Jeff: e

My-Name-Is-Jeff (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+    private val playerPattern = Regex("(?<rank>.+? |§7)?(?<name>\\w+)§r(?<status>§a|§c) ● ")
+    private val party = mutableListOf<PartyMember>()
+
+    //0 = not awaiting, 1 = awaiting 2nd delimiter, 2 = awaiting 1st delimiter
+    private var awaitingMessage = 0
+
+    @SubscribeEvent
+    fun onCommandRun(event: SendChatMessageEvent) {
+        if (!Utils.inSkyblock || !Skytils.config.partyAddons) return
+        if (Utils.equalsOneOf(event.message, "/pl", "/party list", "/p list", "/party l")) {
```
Sychic: Can't you use a set for this?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
Sychic: You can use [regex destructuring](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-match-result/-destructured/) instead of breaking it down like this and then use a let

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -0,0 +1,228 @@
...
+                        PartyMemberType.MEMBER,
+                        status,
+                        rank
+                    )
+                )
+            }
+            event.isCanceled = true
+        } else if (message.startsWith("§cYou are not currently in a party.") && awaitingMessage != 0) {
+            party.clear()
+        } else if (message == "§9§m-----------------------------------------------------§r" && awaitingMessage != 0) {
```
Sychic: Maybe Lily has other opinions about this but we could probably avoid hard coding the message and instead do a contains check for a lot of hyphens. Just in case Hypixel decides to changes things up for no reason.

Sychic (commented):
Not specifically for this pr but maybe we should set up some kind of system to catch messages with delimiters easier.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
My-Name-Is-Jeff: startswith maybe? unless you want a regex

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -38,13 +38,13 @@ object PartyAddons {
...
-    private var awaitingMessage = 0
+    private var awaitingDelimiter = 0
 
     @SubscribeEvent
     fun onCommandRun(event: SendChatMessageEvent) {
-        if (!Utils.inSkyblock || !Skytils.config.partyAddons) return
-        if (Utils.equalsOneOf(event.message, "/pl", "/party list", "/p list", "/party l")) {
-            awaitingMessage = 2
+        if (!Utils.isOnHypixel || !Skytils.config.partyAddons) return
+        if (event.message in setOf("/pl", "/party list", "/p list", "/party l")) {
```
Sychic: This can be changed to a static member to avoid an object allocation

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -53,59 +53,56 @@ object PartyAddons {
...
+                }
             }
             event.isCanceled = true
-        } else if (message.startsWith("§cYou are not currently in a party.") && awaitingMessage != 0) {
+        } else if (message.startsWith("§cYou are not currently in a party.") && awaitingDelimiter != 0) {
             party.clear()
-        } else if (message == "§9§m-----------------------------------------------------§r" && awaitingMessage != 0) {
-            awaitingMessage--
-            if (awaitingMessage == 1 || party.isEmpty()) return
+        } else if (message.startsWith("§9§m----") && awaitingDelimiter != 0) {
```
Sychic: Can we use the unformatted text here? Just to avoid breaking if hypixel ever changes their formatting again.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PartyAddons.kt
```diff
@@ -53,59 +53,56 @@ object PartyAddons {
...
+                    party.add(
+                        PartyMember(
+                            name,
+                            PartyMemberType.MEMBER,
+                            status,
+                            rank
+                        )
                     )
-                )
+                }
```
Sychic: I feel like all of this could be changed into a single block instead of 3 different cases.

Sychic (approved):
Probably could get away with using a map or something to translate the beginning bit to the party member type but that's just nitpicking.

PR #348, opened 2023-03-15
Replace SplashProgress transformer with a mixin

ToxicAven:
This Pull simply replaces the messy transformer to create the custom loading splashes with a nice clean Mixin. In the future, I may look into replacing more of these transformers if it is possible.


My-Name-Is-Jeff:
Mixin was purposely replaced in this commit
https://github.com/Skytils/SkytilsMod/commit/48adf4239755cb089679eea805d1e7a47df49da2

Don't recall why

ToxicAven:
Hmm, doesn't quite make sense, should more testing be done or should this be closed?

(Also great I spent a fair bit of time figuring out how to get it working, glad it was done for me at some point before heh)

PR #355, opened 2023-03-23
Alignment task solver modified

Suyarumi:
I made solver work during maxor and storm
I think this change good for someone do Alignment task early


My-Name-Is-Jeff:
Don't think this is useful for majority of users, there is a secret setting present for the people it is useful for.

Feel free to reopen PR if my think my thought process is incorrect.

PR #367, opened 2023-03-31
Blaze slayer fixes and features

Desco1:
Fixes blaze slayer features being broken and adds 3 new features to blaze slayer.
These are not meant to change how slayer detection or any of the other features work internally to not conflict with the feature/slayer branch. Instead, these are meant to be a temporary solution until SlayerFeatures gets rewritten.

Actual changes:
- Changed timerRegex to properly match all Inferno Demonlord bosses after the demons phase (it also matches demons now, for the Attunement Display feature)
- Updated Slayer Boss Hitbox to also draw around Inferno Demonlord minibosses
- Attunement Display: recolors the Inferno Demonlord or the appropiate demon according to the correct Attunement to use. Unlike the Voidgloom's phase display, the colors are not configurable because I don't think there is a reason since they match the Attunement. Ashen might be too hard to view compared to the rest however
- Ignore pacified blazes: stops rendering blaze entities further from 3 blocks from the player when Smoldering Polarization is active and the Inferno Demonlord is spawned
- Fire pit warning: displays a title and plays a ping when standing in the fire pit that the Inferno Demonlord spawns (that does a billion damage)

Note about the Slayer RNGesus Meter: still broken after the chat message changed. While fixable by keeping track of the open gui, checking if it's a slayer RNG Meter, checking which drop is selected and saving how much is the target XP of the drop, I figured that it'd be easier to save and download the XP required for each drop and use the chat message after changing RNG Meter drop to get it instead.

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
@@ -2453,6 +2453,14 @@ object Config : Vigilant(
     var attunementDisplay = false
 
     @Property(
+        PropertyType.SWITCH, name = "Ignore pacified blazes",
+        description = "Stops rendering faraway blazes when fighting the Inferno Demonlord if Smoldering Polarization is active.\n" +
+                "Do note that you will still be able to interact with them! /skytilsupdatepotioneffects",
+        category = "Slayer", subcategory = "Inferno Demonlord"
+    )
+    var ignorePacifiedBlazes = false
+
+    @Property(
         type = PropertyType.SWITCH, name = "Disable Cooldown Sounds",
         description = "Blocks the sound effect played while an item is on cooldown.",
         category = "Sounds", subcategory = "Abilities"
--- a/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
@@ -2461,6 +2461,13 @@ object Config : Vigilant(
     var ignorePacifiedBlazes = false
 
     @Property(
+        type = PropertyType.SWITCH, name = "Show a warning when in Inferno Demonlord fire",
+        description = "Shows a warning when you are standing on Inferno Demonlord's fire.",
+        category = "Slayer", subcategory = "Inferno Demonlord"
+    )
+    var blazeFireWarning = false
+
+    @Property(
         type = PropertyType.SWITCH, name = "Disable Cooldown Sounds",
         description = "Blocks the sound effect played while an item is on cooldown.",
         category = "Sounds", subcategory = "Abilities"
--- a/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
@@ -2232,6 +2232,16 @@ object Config : Vigilant(
     var voidRNG = 0f
 
     @Property(
+        type = PropertyType.DECIMAL_SLIDER, name = "Current Inferno RNG Meter",
+        description = "Internal value to store current Inferno Demonlord RNG meter",
+        category = "Slayer",
+        decimalPlaces = 1,
+        maxF = 100f,
+        hidden = true
+    )
+    var blazeRNG = 0f
+
+    @Property(
         type = PropertyType.SWITCH, name = "Click to Open Maddox Menu",
         description = "Open chat, then click anywhere on screen to open Maddox Menu.",
         category = "Slayer", subcategory = "Quality of Life"
--- a/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
@@ -2453,7 +2453,7 @@ object Config : Vigilant(
     var attunementDisplay = false
 
     @Property(
-        PropertyType.SWITCH, name = "Ignore pacified blazes",
+        PropertyType.SWITCH, name = "Ignore Pacified Blazes",
         description = "Stops rendering faraway blazes when fighting the Inferno Demonlord if Smoldering Polarization is active.\n" +
                 "Do note that you will still be able to interact with them! /skytilsupdatepotioneffects",
         category = "Slayer", subcategory = "Inferno Demonlord"
@@ -2461,7 +2461,7 @@ object Config : Vigilant(
     var ignorePacifiedBlazes = false
 
     @Property(
-        type = PropertyType.SWITCH, name = "Show a warning when in Inferno Demonlord fire",
+        type = PropertyType.SWITCH, name = "Show a Warning when in Inferno Demonlord Fire",
         description = "Shows a warning when you are standing on Inferno Demonlord's fire.",
         category = "Slayer", subcategory = "Inferno Demonlord"
     )
--- a/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
@@ -2446,6 +2446,13 @@ object Config : Vigilant(
     var totemPing = 0
 
     @Property(
+        PropertyType.SWITCH, name = "Attunement Display",
+        description = "Recolors the Inferno boss and demons depending on the correct dagger attunement.",
+        category = "Slayer", subcategory = "Inferno Demonlord"
+    )
+    var attunementDisplay = false
+
+    @Property(
         type = PropertyType.SWITCH, name = "Disable Cooldown Sounds",
         description = "Blocks the sound effect played while an item is on cooldown.",
         category = "Sounds", subcategory = "Abilities"
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
@@ -249,9 +249,9 @@ object SlayerFeatures : CoroutineScope {
         hasSlayerText = index != -1
         if (!lastTickHasSlayerText && hasSlayerText) {
             sidebarLines.elementAtOrNull(index - 1)?.let {
-                val boss = it.substringBeforeLast(" ")
+                val boss = it.substringBefore(" ")
                 val tier = it.substringAfterLast(" ")
-                expectedMaxHp = BossHealths[boss.substringBefore(" ")]?.get(tier) ?: 0
+                expectedMaxHp = BossHealths[boss]?.get(tier) ?: 0
             }
         }
         slayer?.tick(event)
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
@@ -35,6 +35,7 @@ import gg.skytils.skytilsmod.events.impl.CheckRenderEntityEvent
 import gg.skytils.skytilsmod.events.impl.PacketEvent.ReceiveEvent
 import gg.skytils.skytilsmod.events.impl.RenderHUDEvent
 import gg.skytils.skytilsmod.features.impl.handlers.MayorInfo
+import gg.skytils.skytilsmod.features.impl.handlers.PotionEffectTimers
 import gg.skytils.skytilsmod.utils.*
 import gg.skytils.skytilsmod.utils.NumberUtil.roundToPrecision
 import gg.skytils.skytilsmod.utils.NumberUtil.toRoman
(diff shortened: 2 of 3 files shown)


src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
```diff
@@ -2428,6 +2438,28 @@ object Config : Vigilant(
...
 
+    @Property(
+        PropertyType.SWITCH, name = "Attunement Display",
+        description = "Recolors the Inferno boss and demons depending on the correct dagger attunement.",
+        category = "Slayer", subcategory = "Inferno Demonlord"
+    )
+    var attunementDisplay = false
+
+    @Property(
+        PropertyType.SWITCH, name = "Ignore pacified blazes",
```
My-Name-Is-Jeff: most of the feature names are like capitalized except for words like a, of, etc

src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
```diff
@@ -2428,6 +2438,28 @@ object Config : Vigilant(
...
+        description = "Stops rendering faraway blazes when fighting the Inferno Demonlord if Smoldering Polarization is active.\n" +
+                "Do note that you will still be able to interact with them! /skytilsupdatepotioneffects",
+        category = "Slayer", subcategory = "Inferno Demonlord"
+    )
+    var ignorePacifiedBlazes = false
+
+    @Property(
+        type = PropertyType.SWITCH, name = "Show a warning when in Inferno Demonlord fire",
```
My-Name-Is-Jeff: most of the feature names are like capitalized except for words like a, of, etc

src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/RendererLivingEntityHook.kt
```diff
@@ -50,6 +51,17 @@ fun setColorMultiplier(
                 cir.returnValue = Skytils.config.seraphNormalPhaseColor.withAlpha(169)
             }
         }
+    } else if (Skytils.config.attunementDisplay && Utils.inSkyblock && entity is EntityLiving) {
+        (slayer as? SlayerFeatures.DemonlordSlayer)?.let {
+            if (entity == it.relevantEntity) {
+                entity.hurtTime = 0
+                it.relevantColor?.let {
+                    // Colors might be too hard to see because of the entities textures and colors,
```
My-Name-Is-Jeff, line 59: in this case, could try making alpha higher

My-Name-Is-Jeff (approved):
Thanks for the pull request! I'm not too familiar with blaze slayer so I'll have Sychic check too.

Logically everything makes sense so

src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/RendererLivingEntityHook.kt
Desco1: I have tried multiple alpha values for every attunement and different colors for Ashen specifically, but in the end this configuration seems to be the clearer since the alpha on the color multiplier doesn't quite translate to a clearer value. The animation when an entity is hurt uses 30% alpha in vanilla and it seems to be the clearest the color can get, which corresponds to 169 alpha on the hook. For Ashen I have tried the vanilla §7 and §0 but none of them look better than §8 in my opinion.

src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
```diff
@@ -2428,6 +2438,28 @@ object Config : Vigilant(
...
+    @Property(
+        PropertyType.SWITCH, name = "Attunement Display",
+        description = "Recolors the Inferno boss and demons depending on the correct dagger attunement.",
+        category = "Slayer", subcategory = "Inferno Demonlord"
+    )
+    var attunementDisplay = false
+
+    @Property(
+        PropertyType.SWITCH, name = "Ignore Pacified Blazes",
+        description = "Stops rendering faraway blazes when fighting the Inferno Demonlord if Smoldering Polarization is active.\n" +
```
Sychic, line 2457: Is this supposed to be "faraway"?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
```diff
@@ -231,13 +245,14 @@ object SlayerFeatures : CoroutineScope {
...
         if (!lastTickHasSlayerText && hasSlayerText) {
-            val currentTier =
-                sidebarLines.find { it.startsWith("Voidgloom Seraph") }
-                    ?.substringAfter("Voidgloom Seraph")?.drop(1)
-                    ?: ""
-            expectedMaxHp = BossHealths["Voidgloom"]?.get(currentTier) ?: 0
+            sidebarLines.elementAtOrNull(index - 1)?.let {
+                val boss = it.substringBeforeLast(" ")
+                val tier = it.substringAfterLast(" ")
+                expectedMaxHp = BossHealths[boss.substringBefore(" ")]?.get(tier) ?: 0
```
Sychic: Why not put the `substringBefore` with the declaration?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
```diff
@@ -1160,9 +1192,125 @@ object SlayerFeatures : CoroutineScope {
...
+                        quaziiTimer
+                    } else {
+                        typhoeusTimer
+                    }
+                } else {
+                    timerEntity
+                } ?: return null
+                val attunement = relevantTimer.displayName.unformattedText.substringBefore(" ").stripControlCodes()
+                return attunementColors[attunement]
+            }
```
Sychic, line 1227: How often are these accessed? May be worth handling it somewhere else.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
```diff
@@ -1160,9 +1192,125 @@ object SlayerFeatures : CoroutineScope {
...
+        // Is there a point making a class for the demons and storing the entity and the timer in the same place?
+        var quazii: EntitySkeleton? = null
+        var quaziiTimer: EntityArmorStand? = null
+        var typhoeus: EntityPigZombie? = null
+        var typhoeusTimer: EntityArmorStand? = null
```
Sychic, line 1232: Could create a data class if you need but this works.

Desco1: The class alternative was just in case the initial demon + timer detection failed, to instead detect the armor stand after a small delay from the demon detection, but delaying the TickTask from 5 ticks to 10 has worked consistently without any problems. All of those variables do look kinda ugly but I guess doesn't matter much since the demons don't have many mechanics

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
Desco1: As often as it gets; I haven't had any problems of performance loss but could check for armor stands being renamed from the packet listener to update the relevant entity. Would rather make a few more tests with a profiler but I don't have enough time to make a proper extensive test atm

src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
Desco1: Yes, nearby blazes aren't hidden because the player can still interact with them and also because there might be a bug with minibosses still moving towards the player but not attacking. The range at which the blazes are hidden is just 3 blocks away, but I couldn't find a more appropiate word for that distance

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
Desco1: Forgot to change that after committing, will be fixed after I rework the relevantEntity and relevantColor getters or I do some profiler tests

src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
Sychic: I mean specifically "faraway" as opposed to "far away". Not trying to nitpick but it seems like a typo maybe?

Desco1: Both of them are correct and mean pretty much the same thing, although "faraway" is an adjective and "far away" is an adverb phrase. Could also be rephrased as "...blazes that are far away..."

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
Desco1: After around half an hour of each boss being spawned, neither enderman nor blaze bosses made enough of an impact to appear in the profiler. I haven't noticed any increase on CPU or Memory usage or any FPS drop.
Although a single test isn't enough to decide on the impact of something like this, I think it's more worthwhile to manage all potential bottlenecks in a full rewrite of the file in the feature/slayer branch.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/SlayerFeatures.kt
```diff
@@ -249,9 +249,9 @@ object SlayerFeatures : CoroutineScope {
         hasSlayerText = index != -1
         if (!lastTickHasSlayerText && hasSlayerText) {
             sidebarLines.elementAtOrNull(index - 1)?.let {
-                val boss = it.substringBeforeLast(" ")
```
Sychic: Does this substring the same way it used to? Looks like one was removed.

Desco1: Yes, it grabs the first word from the boss name and properly gets the expected hp, just without the redundant variable assignment. Compared to main branch, the system is used for the "Use Hits to Detect Slayer" feature, but it now works for all slayers instead of only Voidgloom.

Sychic (approved):
LGTM, assuming you've tested it

Desco1:
Before merging there are a few things that could be considered:
- RNG Meter display still broken after chat, which could be fixed easiest in my opinion by downloading the required xp for each drop, a few more details in my first message
- Feature: open RNG Meter menu after a boss if the selected boss has dropped the selected item. Can easily be done by comparing stored xp and sending the /cb command of the message, but not sure if it could somehow be considered a cheat which is why I haven't added it yet.
- Feature: a visual display for every blaze that is affected by the 2x damage from the Moody Grappleshot. Not too useful unless abusing tactical insertion, but might be neat to have? From a bit of testing though, it doesn't seem that there is any easy way to detect when a blaze is being pulled
I don't think either of those are critical enough so merging and then following this pr with another in some future is also fine

Sychic:
>     * RNG Meter display still broken after chat, which could be fixed easiest in my opinion by downloading the required xp for each drop, a few more details in my first message

How did the message change? I haven't done slayer in forever so I have no clue what it is now or when it changed. Did it change from percentage with a bar?

>     * Feature: open RNG Meter menu after a boss if the selected boss has dropped the selected item. Can easily be done by comparing stored xp and sending the /cb command of the message, but not sure if it could somehow be considered a cheat which is why I haven't added it yet.

We could do something similar to the "open maddox batphone" feature where you just click anywhere instead of clicking on a specific bit.

>     * Feature: a visual display for every blaze that is affected by the 2x damage from the Moody Grappleshot. Not too useful unless abusing tactical insertion, but might be neat to have? From a bit of testing though, it doesn't seem that there is any easy way to detect when a blaze is being pulled

We could do something where the blaze itself checks if there's the anvil + the lead nearby (there's got to be some invisible entity that the lead is tied to, no clue what it could be tho).

>       I don't think either of those are critical enough so merging and then following this pr with another in some future is also fine

Also works for me.

Desco1:
> RNG meter chat message

[Reference](https://i.imgur.com/nlXiwds.png). Yes, it now only displays the stored XP, not which drop was selected nor % to it or the next. When the RNG Meter is complete, the drop is shown in chat without magic find, but if it drops by itself, even when selected, the magic find will be shown.
[More reference](https://i.imgur.com/MXb7I4M.png). There is also this chat message shown when a drop is selected from the meter gui, which also shows the slayer it is from. All drops have the same XP for every single player, which is why I think it could be downloaded from server at startup.
Another alternative could be permanently looking for the RNG Meter gui and grabbing the required XP from [there](https://i.imgur.com/k5dlahU.png), but it gets [truncated when actually selected](https://i.imgur.com/PI3AIcs.png), which is pretty annoying
> RNG meter menu like maddox phone

Thought of that, but I didn't want to deal with the consequences of having both messages at once. The Stored XP message could not show when autoslayer is disabled, although I may have just missed it. Could also make it so a different mouse button is a different menu?
> Hooks

When a player uses the hook an S1BPacketEntityAttach gets sent, `leash` is not 0, `vehicleEntityId` is the player who used it and `entityID` is a bat, which most of the times is not in the world yet and its spawn has to be handled in a different event. I've tried recreating vanilla collissions from that bat to blazes but I didn't get any.
I have tried looking for the "snap" effect where some blazes teleport slightly to the hook and checking if blazes have the same velocity as the bat (they do in X and Z, but it is always 0); but I didn't get consistent results for that detection.
Also, the hook 2x damage doesn't last forever, but that can be easily tested compared to the detection

Sychic:
> RNG meter chat message

We should be able to get all the data when the player is actually clicking (though that would take a bit of trust in hypixel to actually receive the click). It's probably just more reliable to download from the data repo and look for when the item is updated to tell if hypixel actually received the click to update the RNG item. 
Though this should probably be separate from all these changes.

> RNG meter menu like maddox phone

Wdym both messages at once? I don't think there's a specific scenario when you would have both the maddox menu and the rng menu (since the maddox one should be entirely manual from using the batphone). We would just keep the most recent `/cb` usage and execute that.

> Hooks

We could use the bat itself to look for blazes near itself (using `Entity#getEntitiesWithinAABBExcludingEntity`) and set a flag in them to tell somewhere else to render them differently. Not entirely version agnostic but not really any better way to do it (though adapting mixins should be pretty simple)

Desco1:
> RNG menu like maddox phone

There isn't a situation where both messages would be active at the same time but I figured it's better to save players from themselves.

In any case, I agree that this PR should not be expanded with more features and instead I'll follow up with the rest of them in some future. Should be ready to merge

PR #368, opened 2023-04-01
Use Coroutines for TickTasks

My-Name-Is-Jeff:
Could completely remove ticktask maybe?

### DIFF
--- a/src/main/java/gg/skytils/skytilsmod/mixins/transformers/MixinMinecraft.java
+++ b/src/main/java/gg/skytils/skytilsmod/mixins/transformers/MixinMinecraft.java
@@ -18,6 +18,7 @@
 
 package gg.skytils.skytilsmod.mixins.transformers;
 
+import com.google.common.util.concurrent.ListenableFuture;
 import gg.skytils.skytilsmod.Skytils;
 import gg.skytils.skytilsmod.core.GuiManager;
 import gg.skytils.skytilsmod.utils.ItemUtil;
@@ -27,6 +28,7 @@ import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.client.gui.ScaledResolution;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
+import org.jetbrains.annotations.NotNull;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
@@ -34,12 +36,16 @@ import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import java.util.Objects;
+import java.util.concurrent.Executor;
 
 @Mixin(Minecraft.class)
-public abstract class MixinMinecraft {
+public abstract class MixinMinecraft implements Executor {
     private final Minecraft $this = (Minecraft) (Object) this;
     @Shadow
     public EntityPlayerSP thePlayer;
+
+    @Shadow
+    public abstract ListenableFuture<Object> addScheduledTask(Runnable runnableToSchedule);
 
     /**
      * Taken from Skyblockcatia under MIT License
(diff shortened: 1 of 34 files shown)


src/main/java/gg/skytils/skytilsmod/mixins/transformers/MixinMinecraft.java
Sychic, line 1: Personally not a fan of mixing in and adding it

src/main/kotlin/gg/skytils/skytilsmod/utils/Utils.kt
```diff
@@ -63,6 +63,7 @@ import java.io.File
 import java.io.IOException
 import java.util.*
 import kotlin.math.floor
+import kotlin.time.Duration.Companion.milliseconds
```
Sychic, line 66: Unnecessary import?

src/main/kotlin/gg/skytils/skytilsmod/utils/tasks/TickTask.kt
```diff
@@ -16,27 +16,50 @@
...
+import kotlinx.coroutines.launch
+import kotlin.coroutines.CoroutineContext
 
-class TickTask<T>(val ticks: Int = 0, val repeats: Boolean = false, register: Boolean = true, val task: () -> T) {
-    var remainingTicks = ticks
+class TickTask<T>(
+    val ticks: Int = 0,
+    val repeats: Boolean = false,
+    register: Boolean = true,
+    val task: (CoroutineContext) -> T
```
Sychic, line 31: Could add a secondary constructor that takes a lambda with no args

src/main/kotlin/gg/skytils/skytilsmod/utils/tasks/TickTask.kt
```diff
@@ -16,27 +16,50 @@
...
+    fun unregister() {
+        job?.cancel() ?: error("Task is not registered!")
+        job = null
+    }
 
-    internal fun complete() = try {
-        callback(task())
+    private fun complete(ctx: CoroutineContext) = try {
+        callback(task(ctx))
     } catch (t: Throwable) {
```
Sychic, line 63: Probably want to catch `CoroutineCancellationException` separately

Sychic (changes requested):
The behavior changes from running every tick to running every game loop

My-Name-Is-Jeff:
> The behavior changes from running every tick to running every game loop

How does coroutine work? I think they accomplish the same thing unless Minecraft is running behind on ticks

Sychic:
> > The behavior changes from running every tick to running every game loop
> 
> How does coroutine work? I think they accomplish the same thing unless Minecraft is running behind on ticks

Well since the coroutine scope has a MC executor as a dispatcher, it's going to queue the tasks there. AFAICT, MC's `addScheduledTask` executes every game loop as opposed to every game tick.

My-Name-Is-Jeff:
> > > The behavior changes from running every tick to running every game loop
> > 
> > 
> > How does coroutine work? I think they accomplish the same thing unless Minecraft is running behind on ticks
> 
> Well since the coroutine scope has a MC executor as a dispatcher, it's going to queue the tasks there. AFAICT, MC's `addScheduledTask` executes every game loop as opposed to every game tick.

Delay behavior should be the same if each tick takes 50ms right? Should be fine because ticks also happen in game loop

Sychic:
> > > > The behavior changes from running every tick to running every game loop
> > > 
> > > 
> > > How does coroutine work? I think they accomplish the same thing unless Minecraft is running behind on ticks
> > 
> > 
> > Well since the coroutine scope has a MC executor as a dispatcher, it's going to queue the tasks there. AFAICT, MC's `addScheduledTask` executes every game loop as opposed to every game tick.
> 
> Delay behavior should be the same if each tick takes 50ms right? Should be fine because ticks also happen in game loop

I think game loops aren't aligned to ticks since game loops are also where the game renders each frame. The delay behavior makes sense but it wouldn't be synced to a tick to begin with.

My-Name-Is-Jeff:
Superseded by 7a741404e010ba19f6b52477341b5720e31c04a7

PR #369, opened 2023-04-02
Fixed Close Button

TakoTheDev:
I hope this is fine, idk why it works this way, but it does


Sychic (commented):
How does this actually do anything? seems like it does the exact same thing as before?

My-Name-Is-Jeff:
> How does this actually do anything? seems like it does the exact same thing as before?

I think both are on different class loaders

TakoTheDev:
> How does this actually do anything? seems like it does the exact same thing as before?

Like idk either, but it works somehow

TakoTheDev:
> > How does this actually do anything? seems like it does the exact same thing as before?
> 
> I think both are on different class loaders

yea seems to be the case, at least to my knowledge

PR #384, opened 2023-05-09
enable petitem confirmation by default

thesefer:
This should be enabled by default. When I installed it, I didn't have pets yet. Today I gave my green bandana pet a farming xp boost by accident. RIP :-(


My-Name-Is-Jeff:
> This should be enabled by default. When I installed it, I didn't have pets yet. Today I gave my green bandana pet a farming xp boost by accident. RIP :-(

sorry for your loss. unfortunately the current stance is that we will not enable features, especially ones that could overlap, by default

PR #387, opened 2023-05-28
Improved interpolation performance in dungeons/slayer hit-box drawing

inglettronald:


Sychic (commented):
Why use the event's coordinates instead of the entity's hitbox?

inglettronald:
This was the first solution we found that improves this issue, and has no performance penalty. It may be possible to interpolate based off of entity hitboxes, but we found that this solution was the easiest that achieved the desired effect.

inglettronald:
This is also an option for rendering stuff if you are willing to change events:
```kt
@SubscribeEvent
    fun onRenderWorldLast(event: RenderWorldLastEvent) {
        
        val players = mc.theWorld.playerEntities.filterNotNull()
        players.forEach {
            val name = it.name ?: return@forEach
            if (name != expectedName) return@forEach
            if (mc.thePlayer.name == name) return@forEach
            val x = it.lastTickPosX + ((it.posX - it.lastTickPosX) * event.partialTicks)
            val y = it.lastTickPosY + ((it.posY - it.lastTickPosY) * event.partialTicks)
            val z = it.lastTickPosZ + ((it.posZ - it.lastTickPosZ) * event.partialTicks)
            WorldRenderUtils.drawCustomBox(
                x - .5,
                1.0,
                y,
                2.0,
                z - .5,
                1.0,
                Color(255, 255, 255, 255),
                3f,
                phase = false
            )
        }
    }
```

This is for a players only, but works for other entities too if you grab the relevant information with `val ents = mc.theWorld.loadedEntityList.filterNotNull()`.

It's also worth noting that my box function doesn't require any player camera interpolation because it's not drawing boxes in any relativity to the player, so this might not work as well for you as it did for me. Worth some testing though. If you just want the easy option for you to implement, there's probably no difference in the end result.

Sychic (commented):
Given the many usages of reconstructing entity bounding boxes, maybe a utility method to directly modify them would be useful?

PR #411, opened 2023-08-30
Update MiningFeatures.kt

Phil-Sieb:
try  to fix moving waypoints


Phil-Sieb:
The Creation of this Pr was an accident, therefore i closed it. It fixes warp cn but allows a death waypoint to spawn from where you ported from if the setting is enabled. The movement on death did not work but was not of importants to  me. (because I dont die xD)

PR #416, opened 2023-09-17
Restructure dungeon chest profit to read drops from croesus chest lore

Harry282:
hi still testing don't merge yet


Harry282:
ok done can merge

PR #418, opened 2023-09-20
added option to disable auto waypoints in crystal hollows

AnanasekE:
added a feature to disable automatic creation of waypoints in the crystal hollows and a config option for that


My-Name-Is-Jeff:
Pretty sure option already exists

PR #419, opened 2023-09-23
Reworked boss timer a bit and many things

AzuredBlue:
+ Fixed Terracotta time in M6 (maybe)
+ Improved Simon Says Solver
+ Fixed noChildLeftBehind (GUI name is different)
+ Reworked M7 Boss Timer (Added a Terminals Split, changed messages to more accepted split messages)
+ Allowed to change box starred mobs color
+ Improved prevent cursor reset to not always reset
= Hide Tooltip in Melody Terminal instead of Maze (Good night maze, not coming back)
= Fixed legion display being wrong sometimes
 - Removed Portal Entry Timer and Hide Pet Nametags (Portal is now instant and you can hide pet nametags in pet settings)

### DIFF
--- a/src/main/java/gg/skytils/skytilsmod/mixins/transformers/util/MixinMouseHelper.java
+++ b/src/main/java/gg/skytils/skytilsmod/mixins/transformers/util/MixinMouseHelper.java
@@ -19,15 +19,21 @@
 package gg.skytils.skytilsmod.mixins.transformers.util;
 
 import com.llamalad7.mixinextras.injector.WrapWithCondition;
-import gg.skytils.skytilsmod.mixins.hooks.util.MouseHelperHookKt;
+import gg.skytils.skytilsmod.mixins.hooks.util.MouseHelperHook;
 import net.minecraft.util.MouseHelper;
+import org.lwjgl.input.Mouse;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(MouseHelper.class)
 public class MixinMouseHelper {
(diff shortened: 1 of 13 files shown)


src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/PetFeatures.kt
```diff
@@ -53,22 +53,6 @@ object PetFeatures {
     private var lastPetLockNotif: Long = 0
     var lastPet: String? = null
 
-
```
My-Name-Is-Jeff, line 56: uhhh I can think of a case where this is useful, cause maybe hypixel hide pet nametags might mess up with some mod feature or something idek

AzuredBlue: doubt a mod uses pet nametags for anything since most people have pets hidden anyways

src/main/java/gg/skytils/skytilsmod/mixins/transformers/util/MixinMouseHelper.java
```diff
@@ -18,16 +18,21 @@
...
 import net.minecraft.util.MouseHelper;
+import org.lwjgl.input.Mouse;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(MouseHelper.class)
 public class MixinMouseHelper {
-    @WrapWithCondition(method = "ungrabMouseCursor", at = @At(value = "INVOKE", target = "Lorg/lwjgl/input/Mouse;setCursorPosition(II)V", remap = false))
```
My-Name-Is-Jeff, line 29: I'm pretty sure wrap with condition is better because if other modifications are made it still works

AzuredBlue: > I'm pretty sure wrap with condition is better because if other modifications are made it still works

no clue how wrap with condition works

AzuredBlue:
i think its done
making simon says boxes more accurately cover the button would be nice but thats too much effort
same with garden features

PR #423, opened 2023-10-09
Blaze puzzle fixes

Harry282:


src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/solvers/BlazeSolver.kt
```diff
@@ -133,6 +134,10 @@ object BlazeSolver {
                 try {
                     val health =
                         blazeName.substringAfter("/").dropLast(1).replace(",", "").toInt()
+                    if (lastKilledBlazeHp != 0 && blazeMode != 0) {
```
My-Name-Is-Jeff, line 137: this method may pose an issue if there are two blazes with the same hp

i still literally cannot see so i'm merging this for now

PR #428, opened 2023-10-17
Feature: Add stacking enchantment stack size

RayDeeUx:
As some items have dungeon stars, I decided to let dungeon stars take priority.

Nullability-related Kotlin syntax is still not quite my forte, so if there's any revisions necessary, let me know.


Sychic:
Unnecessary due to the hud element we provide.

PR #429, opened 2023-10-17
Feature: Abiphone contact capacity

RayDeeUx:
also made sure it wouldn't fire if in neu custom auction house

let me know if my kotlin syntax needs improvement


Sychic:
Unnecessary and doesn't match with skytils's feature set.

PR #430, opened 2023-10-23
Feature(s): Click prevention on a few things

RayDeeUx:
#CLOSED. SEE #439.
- Pickup minion
- Mayor voting
- Wardrobe unequiping
- Abiphone calling

TAB key is a bit unorthodox of a key to hold to bypass the preventions, but as a macOS user I don't trust whatever comes out of Ctrl-click (macOS has some system-specifc "ctrl + click = right click" schtick that I'm not a big fan of). I *was* going to use ALT, but Hypixel seems to use that key pretty often as a stand-in for mouse button clicks, and LMETA triggers the Windows start menu, so TAB it is unless if there's better ideas.


RayDeeUx:
#CLOSED. SEE #439.

PR #431, opened 2023-11-03
Feature: Museum IRL Value on Tooltip

RayDeeUx:
Complements NEU's `/pv` calculations, which are based on unsoulbound value.

Proof of above sentence:

Demo:


RayDeeUx:
Clarification: the Regex is designed as is in case Hypixel decided to move the Coins value into the same line as "your museum value". See https://regex101.com/r/KkfUGQ/1 for reference.

Sychic:
Unnecessary and doesn't match with skytils's feature set.

PR #432, opened 2023-11-07
Add hud text shadow configuration

RectangularObject:
Hooray for a pr on a suggestion made 2 years ago!
image
This replaces the spam text shadow config with an option to change all hud text shadows, and will allow skytils' hud elements to match other mods, such as Apec. Also cleaned up the code a bit using intellij's "Reformat Code" function.

Ideally, I would've liked to implement config options to change each hud element's font shadow individually, but oh well! This'll do for now. Maybe I'll do that whenever I find the motivation to.


RectangularObject:
Reverted the cleanup since it's not really related to the pr, I'm assuming @Sychic or @My-Name-Is-Jeff will clean up the code in a future commit.

Sychic:
Hey, thanks for the PR. While this is a solution to the problem, a preferred solution would be to provide configuration for each gui element separately. This could be achieved by adding a field to the base `GuiElement` class and reading from there. Toggling the option could then be handled by right clicking the element.

RectangularObject:
> Toggling the option could then be handled by right clicking the element.

Huh, I didn't think of this. I'll work on it.

Sychic:
If you do, please try to use an elementa state to contain the value so it's easier to migrate later.

RectangularObject:
An elementa state?

RectangularObject:
Well, there we go. It works, you can go into the configuration file and change every element's text shadow setting individually. Now I just have to figure out how to change that setting in-game using a right or middle click on the element. Is it a bad time to say that I've never really used Kotlin or Java before? Running off of pure intuition here. I'll have to look into how to actually use elementa tomorrow; I don't know much about it. It looks interesting!

src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
```diff
@@ -88,6 +88,14 @@ object Config : Vigilant(
         UDesktop.browse(URI.create("https://discord.gg/skytils"))
     }
 
+    @Property(
```
My-Name-Is-Jeff: Revert

src/main/kotlin/gg/skytils/skytilsmod/utils/Utils.kt
```diff
@@ -218,6 +219,14 @@ object Utils {
 
     fun getKeyDisplayStringSafe(keyCode: Int): String =
         runCatching { GameSettings.getKeyDisplayString(keyCode) }.getOrNull() ?: "Key $keyCode"
+
+    fun getTextShadowSetting(): SmartFontRenderer.TextShadow {
```
My-Name-Is-Jeff: Revert

My-Name-Is-Jeff (changes requested):

My-Name-Is-Jeff:
> Well, there we go. It works, you can go into the configuration file and change every element's text shadow setting individually. Now I just have to figure out how to change that setting in-game using a right or middle click on the element. Is it a bad time to say that I've never really used Kotlin or Java before? Running off of pure intuition here. I'll have to look into how to actually use elementa tomorrow; I don't know much about it. It looks interesting!

https://github.com/Skytils/SkytilsMod/blob/7a741404e010ba19f6b52477341b5720e31c04a7/src/main/kotlin/gg/skytils/skytilsmod/gui/LocationEditGui.kt#L105-L113

RectangularObject:
Cool, it works.
textShadow-preview-optimized

RectangularObject:
Is this still being looked at?

My-Name-Is-Jeff:
> Is this still being looked at?

Forgot about it ngl will review rn

src/main/kotlin/gg/skytils/skytilsmod/core/GuiManager.kt
```diff
@@ -236,13 +238,15 @@ object GuiManager : PersistentSave(File(Skytils.modDir, "guipositions.json")) {
     }
 
     override fun read(reader: Reader) {
-        json.decodeFromString<Map<String, GuiElementLocation>>(reader.readText()).forEach { name, (x, y, scale) ->
+        json.decodeFromString<Map<String, GuiElementLocation>>(reader.readText()).forEach { name, (x, y, scale, textShadow) ->
             val pos = x to y
             GUIPOSITIONS[name] = pos
             GUISCALES[name] = scale
+            GUITEXTSHADOWS[name] = textShadow // TODO: Fix/Work around NullPointerException with old guipositions file
```
My-Name-Is-Jeff: comment resolved?

RectangularObject: No.. This will crash if you have a config file from a previous version.

My-Name-Is-Jeff (approved):

PR #433, opened 2023-11-13
Feature: Missing Trophy Fish tracker (Closes suggestion 3246)

RayDeeUx:
Originally the plan was to scrape SkyCrypt's API (since Hylin is on public archive), but then I re-read the suggestion and figured "hmm, it's still possible to track the types of trophies missing", so here we are:


RayDeeUx:
# CLOSED BY #438

PR #434, opened 2023-11-14
Delete treasure chest opening thing

WcaleNieKot:
delete treasure chest opening thing, as it is no longer needed


My-Name-Is-Jeff:
?

PR #438, opened 2024-01-05
 Feature: Missing Trophy Fish tracker (Closes suggestion 3246) (restores #433 in order to merge into correct branch)

RayDeeUx:
*this PR restores #433 in order to have my commits be merged into the correct branch. original PR description below.*

>Originally the plan was to scrape SkyCrypt's API (since Hylin is on public archive), but then I re-read the suggestion and figured "hmm, it's still possible to track the types of trophies missing", so here we are:
>
>
>
>
>


src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/TrophyFishingProgress.kt
```diff
@@ -0,0 +1,258 @@
...
+    in Trophy Fishing. This is in a separate file instead of being in MiscFeatures.kt
+    as this HUD display relies on multiple Forge events
+    (and so would be very out-of-place if inside MiscFeatures.kt).
+
+*   @author Erymanthus | RayDeeUx
+*/
+
+object TrophyFishingProgress {
+
+    private val trophyFishingProgressDisplay = TrophyFishingProgressDisplay() //do not delete this otherwise the HUD WONT SHOW
```
TakoTheDev, line 62: From what I can tell this val is unused, so you should probably use an [init block](https://github.com/Skytils/SkytilsMod/blob/70575a0aedfa1639133baeb5c78f3b386db7bad0/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/BossHPDisplays.kt#L50)?

RayDeeUx:
manually closing per https://github.com/Skytils/SkytilsMod/commit/b59ec1a463956b4e8cadcd5e70d12d4aa683943d.

PR #441, opened 2024-01-17
Feature: Highlight/Hide croesus chests

Harry282:
Changes:
- Removes shiny prefix from profit calc since there's no item ID for it
- Highlight Croesus run chests based on how many more chests can be opened
  - None opened yet = green
  - One opened and have dungeon chest key = yellow
  - No more can be opened = red or hidden
image

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ChestProfit.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ChestProfit.kt
@@ -20,23 +20,27 @@ package gg.skytils.skytilsmod.features.impl.dungeons
 import gg.essential.api.EssentialAPI
 import gg.essential.universal.UResolution
 import gg.skytils.skytilsmod.Skytils
+import gg.skytils.skytilsmod.Skytils.Companion.mc
 import gg.skytils.skytilsmod.core.structure.GuiElement
 import gg.skytils.skytilsmod.events.impl.GuiContainerEvent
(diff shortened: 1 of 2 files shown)


src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ChestProfit.kt
```diff
@@ -136,7 +164,7 @@ object ChestProfit {
             val enchant = name.substring(name.indexOf("(") + 1, name.indexOf(")"))
             return enchantNameToID(enchant)
         } else {
-            val unformatted = name.stripControlCodes()
+            val unformatted = name.stripControlCodes().replace("Shiny ", "")
```
My-Name-Is-Jeff, line 167: not sure about this one

My-Name-Is-Jeff (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ChestProfit.kt
Harry282: unless there's a different way of doing it, they currently show up as 0 coins in the profit overview in croesus (without clicking into the chest individually)

My-Name-Is-Jeff: actually its dungeons only was worried that they might add an item with shiny in the name

My-Name-Is-Jeff (approved):

PR #444, opened 2024-02-09
feat: add MojangUtil.queueUsernameForUUID

My-Name-Is-Jeff:
Queue usernames for UUID lookup and wait until completed

Run usernames in batch by waiting for at least 10? or max 3 seconds

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
@@ -20,11 +20,15 @@ package gg.skytils.skytilsmod.utils
 
 import gg.essential.lib.caffeine.cache.Cache
 import gg.essential.lib.caffeine.cache.Caffeine
+import gg.skytils.skytilsmod.Skytils
 import gg.skytils.skytilsmod.Skytils.Companion.client
 import io.ktor.client.call.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.http.*
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.launch
 import kotlinx.serialization.Serializable
 import net.minecraft.client.entity.EntityOtherPlayerMP
 import net.minecraftforge.common.MinecraftForge
@@ -34,6 +38,7 @@ import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 import java.util.*
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
+import kotlin.collections.ArrayDeque
 import kotlin.concurrent.fixedRateTimer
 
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
```diff
@@ -65,6 +73,13 @@ object MojangUtil {
         }
     }
 
+    suspend fun queueUsernameForUUID(name: String): UUID? {
+        val username = name.lowercase()
+        usernameQueue.emit(username to System.currentTimeMillis())
+
+        return uuidFlow.first { it.first == username }.second
```
My-Name-Is-Jeff, line 80: if 2 queue the same username what will happen?

src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
```diff
@@ -107,6 +162,19 @@ object MojangUtil {
         fixedRateTimer("Mojang-Fake-Requests-Insert", startAt = Date((System.currentTimeMillis() / 60000) * 60000 + 48000), period = 60_000L) {
             requestCount.set(0)
         }
+
+        Skytils.IO.launch {
+            val usernames = ArrayDeque<Pair<String, Long>>(10)
+            usernameQueue.collect { name ->
+                usernames.add(name)
+                if (usernames.size >= 10 || usernames.isNotEmpty() && System.currentTimeMillis() - usernames.first().second >= 3000L) {
```
My-Name-Is-Jeff, line 170: this condition good?

src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
```diff
@@ -107,6 +162,19 @@ object MojangUtil {
...
+                    getUUIDsFromUsernames(usernames.map { it.first }).forEach {
```
My-Name-Is-Jeff, line 171: would it be better to directly emit the API response?

src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
Sychic: It should still emit. Only `StateFlow` will disregard repeated consecutive elements.

src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
Sychic: If you're trying to create what is effectively a bucket ratelimiter, there are probably better ways to handle this (see [Levelhead](https://github.com/Sk1erLLC/Levelhead/blob/master/src/main/kotlin/club/sk1er/mods/levelhead/core/RateLimiter.kt)). Otherwise, you'd probably want your `usernameQueue` flow to have a replay cache (and buffer) of 10 so you can access queued values more than once. 
Also, It'd probably be nice to add qualifying parenthesis around the second condition because it's not very straightforward from a glance.

src/main/kotlin/gg/skytils/skytilsmod/utils/MojangUtil.kt
```diff
@@ -65,6 +73,13 @@ object MojangUtil {
         }
     }
 
+    suspend fun queueUsernameForUUID(name: String): UUID? {
+        val username = name.lowercase()
+        usernameQueue.emit(username to System.currentTimeMillis())
+
+        return uuidFlow.first { it.first == username }.second
+    }
```
Sychic, line 81: Is there any guarantee that the username will be resolved? Afaict you need to pray someone else calls `MojangUtil#getUsernameFromUUID`.

Sychic (commented):
Overall looks fine but a few concerns.

PR #445, opened 2024-02-09
Show Pet Candies now uses nbt

TakoTheDev:
You guys missed me right?

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
@@ -24,6 +24,7 @@ import gg.skytils.skytilsmod.Skytils
 import gg.skytils.skytilsmod.Skytils.Companion.json
 import gg.skytils.skytilsmod.Skytils.Companion.mc
 import gg.skytils.skytilsmod.core.GuiManager
+import gg.skytils.skytilsmod.utils.SkillUtils.level
 import gg.skytils.skytilsmod.core.structure.GuiElement
 import gg.skytils.skytilsmod.core.tickTimer
 import gg.skytils.skytilsmod.events.impl.GuiContainerEvent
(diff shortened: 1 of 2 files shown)


src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -646,12 +646,11 @@ object ItemFeatures {
...
-        if (Skytils.config.showPetCandies && item.item === Items.skull) { // TODO: Use NBT
-            lore?.forEach { line ->
-                candyPattern.find(line)?.let {
-                    stackTip = it.groups[1]!!.value
-                    return@forEach
-                }
+        if (Skytils.config.showPetCandies && item.item === Items.skull) {
+            val petInfoString = getExtraAttributes(item)?.getString("petInfo")
+            if (!petInfoString.isNullOrBlank()) {
+                val petInfo = json.decodeFromString<PetInfo>(petInfoString)
```
Sychic: Is this tested and does it always work? Try just going threw a few pages on ah and seeing if anything breaks.

TakoTheDev: It worked perfect in my testing, no errors occurred.

TakoTheDev: Actually

TakoTheDev: Okay so, it always works, but it still shows the candy used on maxed out pets. Is that fine?

TakoTheDev: Same with 0/10 Candy used, should I let it show 0 or hide it?

Sychic: > Okay so, it always works, but it still shows the candy used on maxed out pets. Is that fine?

What do you mean by maxed out pets? Level 100?


> Same with 0/10 Candy used, should I let it show 0 or hide it?

Do whatever the old code does.

TakoTheDev: From what I understand level 100 pets that can't be upgraded anymore don't show the Candy used line while sometimes still having the candyUsed nbt

TakoTheDev: image

Sychic: Ah, yeah we might want to make it also hide on maxed pets since it would make for a more continuous experience.

PR #446, opened 2024-02-09
Fix 'drawAllInList' function

TakoTheDev:
At least in the demo renderer it seems to not align correctly when any scaling is applied (Also fixes 3 minor issues: Dungeon Timer height is incorrect, the width and demo renderer for Giant Names don't use same length and the dungeon profit width is missing a +)

**BEFORE**
image

**AFTER**
image
(idk how to fix that very slight space after the heart of the Giant display)


src/main/kotlin/gg/skytils/skytilsmod/utils/RenderUtil.kt
```diff
@@ -863,10 +863,11 @@ object RenderUtil {
     }
 
     fun drawAllInList(element: GuiElement, lines: Collection<String>) {
+        val width = lines.maxOf { ScreenRenderer.fontRenderer.getStringWidth(it) }
```
Sychic: Is there a reason why this is used rather than the scaled width? It doesn't seem to take scaling into account.

TakoTheDev: Imma commit an improvement in a second, for some reason `element.width` (WIDTHWIDTH) is the correct width, `element.scaleWidth` (SCALEDWIDTH) is wrong and can't be bothered to find out why. (CORRECTWIDTH is the string width)
image

TakoTheDev: So tldr: element width is used instead of scaled width since it's wrong for some reason

TakoTheDev:
Also as a side note, optifine seems to break the width calculations?
image
image
(Counts for both old and new editing gui, but can probably not be fixed easily so I'll just leave it like that)

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
```diff
@@ -176,6 +179,10 @@ object DungeonListener {
                         val partyCount = partyCountPattern.find(tabEntries[0].second)?.groupValues?.get(1)?.toIntOrNull()
                         if (partyCount != null) {
                             println("There are $partyCount members in this party")
+                            if (team.size != partyCount) {
+                                println("Clearing team as the party size has changed ${team.size} -> $partyCount")
+                                team.clear()
+                            }
```
Sychic: Why is all this changed? It seems unrelated

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ChestProfit.kt
```diff
@@ -312,7 +312,7 @@ object ChestProfit {
         override val height: Int
             get() = ScreenRenderer.fontRenderer.FONT_HEIGHT * DungeonChest.entries.size
         override val width: Int
-            get() = ScreenRenderer.fontRenderer.getStringWidth("Obsidian Chest: 300M")
+            get() = ScreenRenderer.fontRenderer.getStringWidth("Obsidian Chest: §a+300M")
```
Sychic:
```suggestion
            get() = ScreenRenderer.fontRenderer.getStringWidth("Obsidian Chest: +300M")
```

Sychic (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
TakoTheDev: this is caused by me not merging the latest commit from the branch I think xd

TakoTheDev:
I forgot to add [no ci] ;-;

Sychic (changes requested):
Remove the random newlines

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
Sychic, line 1: This one too

Sychic (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
TakoTheDev: I can't push the random space after the } cause git or intellij ;-;

TakoTheDev:
It had a new line but without the \r\n thing ig?

My-Name-Is-Jeff (approved):

PR #459, opened 2024-03-02
feat: Add CataclysmicMap (Dungeon Map)

My-Name-Is-Jeff:


src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/cataclysmicmap/CataclysmicMap.kt
```diff
@@ -0,0 +1,111 @@
...
+
+        if (DungeonScanner.shouldScan) {
+            DungeonScanner.scan()
+        }
+    }
+
+    @SubscribeEvent
+    fun onWorldLoad(event: WorldEvent.Load) {
+        DungeonInfo.reset()
+        MapUtils.calibrated = false
```
TakoTheDev: Why not just use the reset function?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/cataclysmicmap/core/CataclysmicMapConfig.kt
```diff
@@ -0,0 +1,411 @@
...
+        description = "Rotates map to follow the player.",
+        category = "Map",
+        subcategory = "Toggle"
+    )
+    var mapRotate = false
+
+    @Property(
+        name = "Center Map",
+        type = PropertyType.SWITCH,
+        description = "Centers the map on the player if Rotate Map is enabled.",
```
TakoTheDev, line 53: Should probably depend on Rotate map right?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/cataclysmicmap/core/CataclysmicMapConfig.kt
```diff
@@ -0,0 +1,411 @@
...
+        category = "Map",
+        subcategory = "Toggle"
+    )
+    var mapCenter = false
+
+    @Property(
+        name = "Dynamic Rotate",
+        type = PropertyType.SWITCH,
+        description = "Keeps the entrance room at the bottom. Does not work with rotate map.",
```
TakoTheDev, line 62: Should probably be in a selector list with rotate map?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/cataclysmicmap/core/CataclysmicMapConfig.kt
TakoTheDev: Actually you can probably only do one of these suggestions? idk

TakoTheDev:
image
image
It seems one player is always missing and instead rendered on top of another player or smth?

And the names should probably not rotate with the head
image

Also while a door is opening it shows the no key highlight, possibly make it not show at all once you open it?

image
If you know how many secrets are done in a room, maybe show it like 3/4 or ?/4 if not known?

image
Maybe show checkmark for yellow room?

image
room has no secrets but is done, maybe show some indicator for that in the secrets replace checkmark setting?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/cataclysmicmap/core/DungeonMapPlayer.kt
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.util.ResourceLocation
+
+data class DungeonMapPlayer(val teammate: DungeonListener.DungeonTeammate, val skin: ResourceLocation) {
+    var mapX = 0
+    var mapZ = 0
+    var yaw = 0f
+
+    /** Has information from player entity been loaded */
+    var playerLoaded = false
+    // TODO: fix wrong icon count when players die
```
My-Name-Is-Jeff: needs research

TakoTheDev:
aware that it is a temp commit, but still why not
image

TakoTheDev:
Looks weird
---
image

Looks weird
---
image

Dynamic Rotate when the Entrance is at the top (likely happens when it is to the side as well):
---
image

> Also while no screenshot of it rn, I still don't think the names should rotate with the head.

Oh also, if you're out of render distance, while someone is blood rushing, the door highlights stay until you load the chunks(?) in, which is not that bad of an issue but exists:
---
image

PR #463, opened 2024-03-10
add: Blaze and Vampire slayers to st pv

Fix3dll:
**Slayers:**
slayer

**Heads: (Before -> After)**
beforeAfter

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/gui/profile/components/ItemComponent.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/gui/profile/components/ItemComponent.kt
@@ -38,7 +38,7 @@ class ItemComponent(val state: State<ItemStack>) : UIComponent() {
         super.draw(matrixStack)
         matrixStack.push()
         matrixStack.translate(getLeft(), getTop(), 100f)
-        matrixStack.scale(getWidth() / 16f, getHeight() / 16f, 0f)
+        //matrixStack.scale(getWidth() / 16f, getHeight() / 16f, 0f)
         UGraphics.color4f(1f, 1f, 1f, 1f)
         matrixStack.runWithGlobalState {
             val item = state.get()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/gg/skytils/skytilsmod/gui/profile/components/ItemComponent.kt
```diff
@@ -38,7 +38,7 @@ class ItemComponent(val state: State<ItemStack>) : UIComponent() {
         super.draw(matrixStack)
         matrixStack.push()
         matrixStack.translate(getLeft(), getTop(), 100f)
-        matrixStack.scale(getWidth() / 16f, getHeight() / 16f, 0f)
```
Sychic, line 41: Is this change necessary? It should adjust the skull to be the appropriate size for an arbitrarily sized `ItemComponent`

Sychic (approved):
Just one question, otherwise LGTM.

src/main/kotlin/gg/skytils/skytilsmod/gui/profile/components/ItemComponent.kt
Fix3dll: This scaling breaking skull textures. I didnt see any problem in scaling when i tried it without this. (as seen in the images above) Thats why i didnt think there was a need to alternative scaling. Another e.g. using ItemComponent is
**SuperboomToast (Old -> New)**
toast

Sychic: Those all used 16 pixel scalings, could you see if it still works with something like 20 pixels or larger like 32?

Fix3dll: `matrixStack.scale(getWidth() / 32f, getHeight() / 32f, 0f)`
image
`matrixStack.scale(getWidth() / 20f, getHeight() / 20f, 0f)`
image
`matrixStack.scale(getWidth() / 4f, getHeight() / 4f, 0f)`
image
😂

Sychic: Hm very interesting. I'll have to investigate this further later.

PR #467, opened 2024-03-17
feat: show item quality

TakoTheDev:
image
image

Do I like generate the l18n stuff myself or?

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
@@ -421,10 +421,11 @@ object ItemFeatures {
             val tier = extraAttr.getInteger("item_tier")
 
             if (boost > 0 && tier > 0) {
-                val isMasterMode = requirementPattern.matchEntire(extraAttr.getString("dungeon_skill_req"))?.let {
-                    val level = it.groupValues[1].toInt()
-                    level > 24
-                }
+                val isMasterMode =
+                    requirementPattern
+                        .matchEntire(
+                            extraAttr.getString("dungeon_skill_req")
+                        )?.groupValues?.get(1)?.toIntOrNull()?.let { it > 24 }
 
                 val floor = when (isMasterMode) {
                     true -> "§4M${tier - 3}"
@@ -433,9 +434,9 @@ object ItemFeatures {
                 }
 
                 val color = when {
-                    boost < 17 -> "§c"
-                    boost < 33 -> "§e"
-                    boost < 49 -> "§a"
+                    boost <= 17 -> "§c"
+                    boost <= 33 -> "§e"
+                    boost <= 49 -> "§a"
                     else -> "§b"
                 }
 
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
@@ -420,12 +420,12 @@ object ItemFeatures {
             val boost = extraAttr.getInteger("baseStatBoostPercentage")
             val tier = extraAttr.getInteger("item_tier")
 
-            val isMasterMode = requirementPattern.matchEntire(extraAttr.getString("dungeon_skill_req"))?.let {
-                val level = it.groups["level"]?.value?.toIntOrNull() ?: return@let null
-                level > 24
-            }
-
             if (boost > 0 && tier > 0) {
+                val isMasterMode = requirementPattern.matchEntire(extraAttr.getString("dungeon_skill_req"))?.let {
+                    val level = it.groups["level"]?.value?.toIntOrNull() ?: return@let null
+                    level > 24
+                }
+
                 val floor = when (isMasterMode) {
                     true -> "§4M${tier - 3}"
                     false -> "§aF$tier"
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
@@ -420,10 +420,7 @@ object ItemFeatures {
             val tier = extraAttr.getInteger("item_tier")
 
             if (boost > 0 && tier > 0) {
-                event.toolTip.addAll(
-                    event.toolTip.size - 1,
-                    listOf("§6Base Stat Boost: §b$boost%", "§6Item Tier: §b$tier")
-                )
+                event.toolTip.addAll(listOf("§6Base Stat Boost: §b$boost%", "§6Item Tier: §b$tier"))
             }
         }
 
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
@@ -88,18 +88,7 @@ object ItemFeatures {
 
     private val headPattern =
         Regex("(?:DIAMOND|GOLD)_(?:(BONZO)|(SCARF)|(PROFESSOR)|(THORN)|(LIVID)|(SADAN)|(NECRON))_HEAD")
-    private val obtainedFromFloor = mapOf(
-        1 to "§aF1",
-        2 to "§aF2",
-        3 to "§aF3",
-        4 to "§aF4§7/§4M1",
-        5 to "§aF5§7/§4M2",
-        6 to "§aF6§7/§4M3",
-        7 to "§aF7§7/§4M4",
-        8 to "§4M5",
-        9 to "§4M6",
-        10 to "§4M7",
-    )
+    private val requirementPattern = Regex("CATACOMBS:(?<level>\\d+)")
 
     // TODO: it is possible for 2 items to have the same name but different material
     val itemIdToNameLookup = hashMapOf<String, String>()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -414,6 +414,16 @@ object ItemFeatures {
                 }"
             })
         }
+
+        if (Skytils.config.showItemQuality && extraAttr != null) {
+            val boost = extraAttr.getInteger("baseStatBoostPercentage")
+            val tier = extraAttr.getInteger("item_tier")
+
+            if (boost > 0 && tier > 0) {
+                event.toolTip.addAll(listOf("§6Base Stat Boost: §b$boost%", "§6Item Tier: §b$tier"))
```
AzuredBlue: I don't think "item_tier" works well enough. Doesn't work well with master mode (Example, item_tier in Master Floor 3 is 6, same as floor 6. Floor 7 is 7, same as Master Floor 4.

Instead, dungeon_skill_req is the same as the requirement to enter the floor it dropped from and it would work for differentiating between normal floors and master mode. 
Also instead of adding "Item Tier:" and the tier i think you can just put that between parenthesis after the stat boost

TakoTheDev: I mean knowing whether it came from F7 or M4 doesn't change anything does it? Since the boost both of them give is the same is it not?

AzuredBlue: > I mean knowing whether it came from F7 or M4 doesn't change anything does it? Since the boost both of them give is the same is it not?

Not sure (since you would need 2 armor pieces with same stat boost to check), but imo its better to do it that way if its going to support master mode anyways, since it avoids confusing hypixel values like "item_tier" 10 being M7 or 8 being M5

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -88,6 +88,7 @@ object ItemFeatures {
 
     private val headPattern =
         Regex("(?:DIAMOND|GOLD)_(?:(BONZO)|(SCARF)|(PROFESSOR)|(THORN)|(LIVID)|(SADAN)|(NECRON))_HEAD")
+    private val requirementPattern = Regex("CATACOMBS:(?<level>\\d+)")
```
TakoTheDev, line 91: Should I do this or (it.startsWith("CATACOMBS:") it.subString(0, idk like 10).toIntOrNull()

Sychic: Regex is better IMO.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -414,6 +415,27 @@ object ItemFeatures {
...
+            if (boost > 0 && tier > 0) {
+                val isMasterMode = requirementPattern.matchEntire(extraAttr.getString("dungeon_skill_req"))?.let {
+                    val level = it.groups["level"]?.value?.toIntOrNull() ?: return@let null
+                    level > 24
+                }
+
+                val floor = when (isMasterMode) {
+                    true -> "§4M${tier - 3}"
+                    false -> "§aF$tier"
+                    null -> "§b$tier"
```
Sychic: What is this case for?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -414,6 +415,27 @@ object ItemFeatures {
...
+                }
+
+                event.toolTip.add("§6Quality: §b$boost% §7($floor§7)")
```
Sychic: Maybe we could associate colors to certain quality ranges? (i.e going from red to green, 0-16% red, 17-33% yellow, 34-49% green, 50% aqua)

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -414,6 +415,27 @@ object ItemFeatures {
...
             })
         }
+
+        if (Skytils.config.showItemQuality && extraAttr != null) {
+            val boost = extraAttr.getInteger("baseStatBoostPercentage")
+            val tier = extraAttr.getInteger("item_tier")
+
+            if (boost > 0 && tier > 0) {
+                val isMasterMode = requirementPattern.matchEntire(extraAttr.getString("dungeon_skill_req"))?.let {
+                    val level = it.groups["level"]?.value?.toIntOrNull() ?: return@let null
```
Sychic: You can directly use `MatchResult#groupValues` here to avoid getting the individual group.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
AzuredBlue: aqua as the best seems really weird, maybe red-orange-yellow-green would be better (although i dont think people care about lower qualities than 50%?)

Sychic: I just used aqua as an example because it's what's currently used. Though having a visually distinct color for best would probably be best as it allows for easier differentiation at a glance.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
TakoTheDev: kk will continue to use Regex in future code as well then

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
TakoTheDev: If they change the dungeon_skill_req String the Regex won't match (Or if the tag is not there at all, tho idk if that can happen) and with that you don't know whether it came from master mode or not so just default to the number representation

Sychic: Ah fair enough, I would prefer the case to be `else` instead just to make clear that it will catch all other values (even though a `Boolean?` has 3 states)

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
TakoTheDev: I feel like aqua fill fit nicely with the other tooltips added at the bottom so I'll use that as the top quality

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -414,6 +415,34 @@ object ItemFeatures {
...
+
+        if (Skytils.config.showItemQuality && extraAttr != null) {
+            val boost = extraAttr.getInteger("baseStatBoostPercentage")
+            val tier = extraAttr.getInteger("item_tier")
+
+            if (boost > 0 && tier > 0) {
+                val isMasterMode = requirementPattern.matchEntire(extraAttr.getString("dungeon_skill_req"))?.let {
+                    val level = it.groupValues[1].toInt()
+                    level > 24
+                }
```
Sychic:
```suggestion
                val isMasterMode = 
                    requirementPattern
                        .matchEntire(
                            extraAttr.getString("dungeon_skill_req")
                        )?.groupValues?.get(1)?.toIntOrNull()?.compareTo(24) != 1
```

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -414,6 +415,34 @@ object ItemFeatures {
...
+                    true -> "§4M${tier - 3}"
+                    false -> "§aF$tier"
+                    else -> "§b$tier"
+                }
+
+                val color = when {
+                    boost < 17 -> "§c"
+                    boost < 33 -> "§e"
+                    boost < 49 -> "§a"
+                    else -> "§b"
```
Sychic, line 440: This includes both 49% and 50%

Sychic (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
TakoTheDev: Oh true

src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
TakoTheDev: ig I could do that, however if they ever change anything about it, it'll say stuff like F10, is that fine?

Sychic: What

Sychic:
```suggestion
                val isMasterMode =
                    requirementPattern
                        .matchEntire(
                            extraAttr.getString("dungeon_skill_req")
                        )?.groupValues?.get(1)?.toIntOrNull()?.let { it > 24 }
```

Sychic: https://discord.com/channels/807302538558308352/807302538558308355/1222183936558366790

Sychic (approved):

PR #471, opened 2024-03-23
fix: ice fill 7x7 taking a while to show up

TakoTheDev:
Tbh I mainly was just testing around with this and at some point realized I fixed the issue (Also at least imo the code is now better).


jani270:
🙏

Sychic (approved):
The root of the issue is that the current solver method is basically random and is not deterministic, ideally this should be rewritten in the future but this looks good enough for now.

PR #475, opened 2024-03-26
fix: sweat mode can't be disabled

TakoTheDev:


Sychic:
Superceded by 96c01dd

PR #476, opened 2024-03-26
fix: always show item highlight

TakoTheDev:
2 issues: both in line 603
image
It basically only modified the check in line 595 before, meaning the opacity was 0 * 256 / 10 which is still 0 and stopped it from rendering.
Second issue is that 1 is only barely visible, so 10 is used instead, as that is the max needed basically.


My-Name-Is-Jeff:
> 2 issues: both in line 603 image It basically only modified the check in line 595 before, meaning the opacity was 0 * 256 / 10 which is still 0 and stopped it from rendering. Second issue is that 1 is only barely visible, so 10 is used instead, as that is the max needed basically.

iirc this class is replaced by optifine? did you check in game

TakoTheDev:
Oh yea optifine also messes with it

CuzImClicks (approved):

PR #478, opened 2024-03-27
fix: Actually fix all ice fill variants

TakoTheDev:
This is just a general rewrite which is not ready for review, just doing this so other people can easily test it.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -34,7 +34,7 @@ plugins {
     signing
 }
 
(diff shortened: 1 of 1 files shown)


TakoTheDev:
Not entirely ready, but ig if you want to look over you can

TakoTheDev:
image
It works, pls pull/review, ty

Edit: Can't be bothered to add the precalced paths since it's fast enough anyway

PR #482, opened 2024-04-28
Major fixes of zh-tw localization

wbxshiori:
- Fixed many mistranslations and unsmooth sentences.
- For **readability**, certain **unofficial** translations (e.g. "僕從" for Minion, "爆炸" for Implosion Skill) were replaced with official English terms/names in singular forms.
- For easier understanding, some translations of common words now have parentheses with their original English terms/names (e.g. "神話儀式 (Mythological Ritual)").


wbxshiori:
I wonder why this pull request has not been proceeded or even discussed for this long.
A very important point is that zh-tw localization should not be simple char-to-char replacements of zh-cn. There are many terms and word usages different to zh-cn (e.g. Enderman is "終界使者" in zh-tw, not "末影人").

PR #489, opened 2024-07-11
Dungeon Score improvements

Desco1:
A revamp of the Dungeon Score Estimate which displays new useful information to the player (min and could skip secrets) and standarizes all information shown. The PR also includes fixes for a few dungeon-related bugs: sending "Blaze Done" too early, rendering the current dungeon chest profit on top of the profit of the rest of the chests, and croesus chest profit (sometimes) assuming starred items, such as shadow assassin boots; reset puzzles not correctly updating score.

About Dungeon Score Estimate, it's easier to follow along with the [before](https://github.com/user-attachments/assets/36e551e7-826f-47db-9e9f-63472189be87) and [after](https://imgur.com/cvEUDUH).
The **Dungeon Status** block is now part of one setting, "Show Dungeon Status", instead of being bundled with the now removed "Minimized Dungeon Score Estimate" (when disabled). This new setting also adds the blank line separating it and the "Dungeon Score" block.
The **Deaths** line is the same, except the number is green if there are no deaths.
"Missing Puzzles" and "Failed Puzzles" have been condensed into **Puzzles**. The number shown is green when all puzzles are completed, yellow when there are puzzles left to complete, and red when there are failed puzzles.
**Secrets** has gotten extra text indicating "min" and "could skip" secrets. "min" shows the remaining amount of secrets that are needed for an S+ in this current run, assuming all rooms and puzzles are going to be cleared (even if they aren't in the moment). "Could skip" secrets is similar, except it assumes the party is going to get 5 crypts and mimic, on the floors it exists. Both of these numbers update with the current clear speed score, deaths, failed puzzles, paul bonus score; and properly display if S+ is no longer possible due to these penalties.
**Crypts** color is now red, green if the party has gotten over 5 of them. The line doesn't display the total amount of crypts gotten, only up to "5", to simplify the information that actually matters to score.
**Mimic** now also shows on floors where there isn't a mimic, showing a gray dash instead of a red cross or a green checkmark. This is to keep vertical size consistent, so the player doesn't have an inconsistent configuration.

The Score block has been renamed to **Dungeon Score** to keep consistency with the previous block. The new "Dungeon Score Breakdown" setting acts like the previous "Minimized Dungeon Score Estimate" setting, which shows or [hides](https://i.imgur.com/X3g2wZI.png) all of the categories of the dungeon score. Otherwise, this entire block has gotten no changes.

Internally, the ScoreCalculation class has been refactored to MAYBE improve readability a bit; all the variables that are read from the dungeon are at the top and the calculations made with them are below. Two more vars have been added to handle the Puzzle line, puzzleCount and completedPuzzleCount; as well as another regex to handle reset puzzles.


LappySheep:
aeiouaeiouaeiouaeiouaeiou

i agree

PR #490, opened 2024-07-15
Dungeon Score improvements (on another branch)

Desco1:
Same thing as https://github.com/Skytils/SkytilsMod/pull/489, except the score calc class was also migrated to the new event system.

Wanted to add a feature to mark Blaze Puzzle as complete for the score calc (and potentially catlas aswell) when the solver detected 10 blaze deaths (or when read from party chat) but figured that that may aswell wait until most players aren't on a version with a bugged "Blaze Done" party chat message.


mod/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
```diff
@@ -449,14 +449,18 @@ object Config : Vigilant(
...
 
     @Property(
-        type = PropertyType.SWITCH, name = "Minimized Dungeon Score Estimate",
-        description = "Only shows the dungeon score.",
-        category = "Dungeons", subcategory = "Score Calculation",
-        i18nName = "skytils.config.dungeons.score_calculation.minimized_dungeon_score_estimate",
-        i18nCategory = "skytils.config.dungeons",
-        i18nSubcategory = "skytils.config.dungeons.score_calculation"
+        type = PropertyType.SWITCH, name = "Show Dungeon Status",
+        description = "Shows information about the current dungeon run that affects its score.",
```
Sychic: Since this PR removes "minimized score estimate", it might be useful to mention that this toggle is effectively the opposite of it. Maybe something along the lines of "Disabling this will minimize the dungeon score"

mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
```diff
@@ -260,55 +346,100 @@ object ScoreCalculation {
...
+                }}${crypts.get().coerceAtMost(5)}")
+                ScoreCalculationElement.text.add("• §eMimic: §l${if ((dungeonFloorNumber ?: 0) >= 6) {
+                    if (mimicKilled.get()) {
+                        "§a✔"
+                    } else {
+                        "§c✘"
+                    }
+                } else {
+                    "§7-"
+                }}")
```
Sychic: Perhaps it would make more sense to expand everything out into multiline string templates to avoid the awkward formatting here. In particular, the if/else statements being expanded while the rest of the string template is not feels weird and also the prevalence of using the same colors for a simple boolean which could maybe be replaced with a helper method.

mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
```diff
@@ -80,11 +82,59 @@ object ScoreCalculation {
...
+    var mimicKilled = BasicState(false)
+    private var isPaul = BasicState(false)
+
+    init {
+        tickTimer(5, repeats = true) {
+            isPaul.set(
+                (MayorInfo.currentMayor == "Paul" && MayorInfo.mayorPerks.contains("EZPZ")) || MayorInfo.jerryMayor?.name == "Paul"
+            )
+        }
+    }
```
Sychic, line 130: This shouldn't change that often so it should be on at least a 20 tick delay in checking. Probably more around 30 seconds ish would make sense.

Sychic (approved):

Sychic:
oh huh i meant to request changes, not approve

mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
Desco1: This was just moved and not changed with the pr, it could be delayed to 60 seconds since that's the delay of the MayorInfo timer

mod/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
Desco1: Unsure about specifying for a deprecated setting, but yeah the settings are the exact opposite, could maybe be convenient to flip them as "Hide Dungeon Status/Score Breakdown" instead. Also, the previous setting was disabled by default, maybe the new ones (if not changed from "Show") could be enabled to keep default functionality?

mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
Desco1: Kinda find it awkward to have only half of the lines on the template, and moving puzzles and secrets outside of it, but it did improve readability a lot, probably worth settling with [it](https://i.imgur.com/iMTbu6A.png) either way?

Desco1:
I'm not sure exactly what I just did but the commit amount doubled? Not sure if it was because of having approved it, but there's a bunch of duplicate commits now, hopefully not causing (a lot of) issues

mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/ScoreCalculation.kt
Sychic: Yeah that looks better

mod/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
Sychic: Maybe, I meant mostly for users who are used to the previous settings/names.

PR #491, opened 2024-07-15
fix: remove limbo command

MicrocontrollersDev:
this method no longer works and now just breaks hypixel's built-in command


Sychic:
Hypixel's limbo command can still be used by using `/server:limbo`. 1.x is considered EOL so this is not planned to be merged.

PR #495, opened 2024-08-24
fix: ice fill solves slowly

TakoTheDev:
ice fill go brrrr

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/solvers/IceFillSolver.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/solvers/IceFillSolver.kt
@@ -17,6 +17,8 @@
  */
 package gg.skytils.skytilsmod.features.impl.dungeons.solvers
 
+import gg.essential.universal.UChat
+import gg.essential.universal.UKeyboard
 import gg.essential.universal.UMatrixStack
 import gg.skytils.skytilsmod.Skytils
 import gg.skytils.skytilsmod.Skytils.Companion.mc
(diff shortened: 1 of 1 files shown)


MX1D (approved):
tako good, me approve, you add

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/solvers/IceFillSolver.kt
```diff
@@ -51,50 +54,65 @@ object IceFillSolver {
...
+                                                chestCenter.down(3).offset(direction.opposite, 10),
+                                            )
+                                            val ends = Triple(
+                                                //three: -29 70 -89
+                                                starts.first.offset(direction, 3),
+                                                //five: -23 71 -89
+                                                starts.second.offset(direction, 5),
+                                                //seven: -14 72 -89
+                                                starts.third.offset(direction, 7),
+                                            )
```
Sychic, line 106: What's the point of hard coding the beginning and ends like this? Does it actually improve the performance significantly?

Sychic (commented):
Overall looks fine but maybe it's no longer necessary to have the abstraction for representing a puzzle when it's just individual points that are drawn.

src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/solvers/IceFillSolver.kt
TakoTheDev: Since it's been months when I originally added it, idk exactly how much the performance increased, however it was definitely noticable from what I remember. As before it scanned a 51x51x3 area, so about 7803 blocks and that times 2 (So 15606 blocks), since it also scans all the blocks below it. You could probably optimize the original code enough for it to be usable but I couldn't be bothered.

Sychic: Alright, just checking since hard coding can mean that updates will break it in the future. Personally I'd prefer scanning (definitely with a more optimized area) over hard coding but if it works, that's fine.

PR #499, opened 2024-09-13
feat: Run breakdown

TakoTheDev:
### Ugly design, open for suggestions!
image

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
@@ -38,6 +38,7 @@ import gg.skytils.skytilsmod.features.impl.dungeons.ScoreCalculation
 import gg.skytils.skytilsmod.features.impl.dungeons.catlas.core.DungeonMapPlayer
 import gg.skytils.skytilsmod.features.impl.dungeons.catlas.core.map.Room
 import gg.skytils.skytilsmod.features.impl.dungeons.catlas.core.map.RoomType
+import gg.skytils.skytilsmod.features.impl.dungeons.catlas.core.map.UniqueRoom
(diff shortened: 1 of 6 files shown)


TakoTheDev:
no idea why it decided to delete `src/main/kotlin/gg/skytils/skytilsws/shared/structs/BoundingBox.kt` and `src/main/kotlin/gg/skytils/skytilsws/shared/structs/CHWaypointType.kt`, ig I'll revert and push that once I get the complaints from lily

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
```diff
@@ -150,7 +151,7 @@ object DungeonListener {
                             if (room.foundSecrets != sec) {
                                 room.foundSecrets = sec
                                 if (team.size > 1)
-                                    WSClient.sendPacketAsync(C2SPacketDungeonRoomSecret(SBInfo.server ?: return@setFoundSecrets, room.mainRoom.data.name, sec))
+                                    WSClient.sendPacketAsync(C2SPacketDungeonRoomSecret(SBInfo.server ?: return@setFoundSecrets, room.mainRoom.data.name, sec, mc.thePlayer.name))
```
My-Name-Is-Jeff: what is the point of this? two players can have the same update at the same time and can be in the same room at the same time

additionally the server is aware of the sender of the packet

My-Name-Is-Jeff (changes requested):

TakoTheDev:
I can't really test rooms and secrets, since I can't keep a websocket connection for some reason, so would be nice if someone else could test it <3 (Even on the version in Releases, I think it's internet issue)

My-Name-Is-Jeff:
> I can't really test rooms and secrets, since I can't keep a websocket connection for some reason, so would be nice if someone else could test it <3 (Even on the version in Releases, I think it's internet issue)

make sure you did not modify packets. the server cannot parse modified packets due to protobuf

PR #502, opened 2024-09-20
fix: thread starvation from non-constant time method call

My-Name-Is-Jeff:
checking size/not empty on ConcurrentLinkedQueue is not constant time and requires either looping or grabbing first element

supersedes #501

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
@@ -32,6 +32,7 @@ import gg.skytils.skytilsmod.core.API
 import gg.skytils.skytilsmod.core.tickTimer
 import gg.skytils.skytilsmod.events.impl.MainReceivePacketEvent
 import gg.skytils.skytilsmod.events.impl.skyblock.DungeonEvent
+import gg.skytils.skytilsmod.events.impl.skyblock.LocationChangeEvent
 import gg.skytils.skytilsmod.features.impl.dungeons.DungeonFeatures
 import gg.skytils.skytilsmod.features.impl.dungeons.DungeonTimer
 import gg.skytils.skytilsmod.features.impl.dungeons.ScoreCalculation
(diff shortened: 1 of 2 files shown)


src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
```diff
@@ -125,10 +130,23 @@ object DungeonListener {
         missingPuzzles.clear()
         completedPuzzles.clear()
         teamCached.clear()
-        outboundRoomQueue.clear()
+        printDevMessage("closed room queue world load", "dungeonws")
+        outboundRoomQueue.close()
+        outboundRoomQueueTask?.cancel()
```
Sychic: Not sure if this does exactly what you expect it to since this will simply wait until it's fully finished executing before cancelling.

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
```diff
@@ -125,10 +130,23 @@ object DungeonListener {
         missingPuzzles.clear()
         completedPuzzles.clear()
         teamCached.clear()
-        outboundRoomQueue.clear()
+        printDevMessage("closed room queue world load", "dungeonws")
+        outboundRoomQueue.close()
```
Sychic: Channel should be cancelled instead of closing since it cancels the receiver immediately too.

src/main/kotlin/gg/skytils/skytilsmod/listeners/DungeonListener.kt
```diff
@@ -222,11 +241,15 @@ object DungeonListener {
                                     entranceLoc = entrance.mainRoom.z * entrance.mainRoom.x
                                 ))
                                 while (DungeonTimer.dungeonStartTime != -1L) {
-                                    while (outboundRoomQueue.isNotEmpty()) {
-                                        val packet = outboundRoomQueue.poll() ?: continue
+                                    for (packet in outboundRoomQueue) {
```
Sychic, line 242: I'm not sure about the underlying implementation of iterating over a channel but is the issue really that `isNotEmpty` is not constant time? i.e. is this suspending rather than blocking when waiting for a new element?

My-Name-Is-Jeff: > I'm not sure about the underlying implementation of iterating over a channel but is the issue really that `isNotEmpty` is not constant time? i.e. is this suspending rather than blocking when waiting for a new element?

iterating over a channel should suspend the coroutine while it waits for a new element

PR #503, opened 2024-09-24
fix: support multi-line trivia

TakoTheDev:
This works with the one-liners, however I haven't gotten the new question so idk if it actually fixes it.

Once I or someone else confirms it works for that question, I'll remove the data url and the printlns.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -34,7 +34,7 @@ plugins {
     signing
 }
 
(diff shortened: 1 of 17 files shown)


TakoTheDev:
link to the file if anyone wants to try: https://github.com/Skytils/SkytilsMod/actions/runs/11017599364

PR #504, opened 2024-09-25
Update Pickonimbus 2000 max durability

lunaynx:


src/main/kotlin/gg/skytils/skytilsmod/features/impl/misc/ItemFeatures.kt
```diff
@@ -653,11 +654,19 @@ object ItemFeatures {
...
             }
             if (extraAttributes.hasKey("pickonimbus_durability")) {
-                RenderUtil.drawDurabilityBar(
-                    event.x,
-                    event.y,
-                    1 - extraAttributes.getInteger("pickonimbus_durability") / 5000.0
-                )
+                val durability = extraAttributes.getInteger("pickonimbus_durability")
+                /*
+                Old Pickonimbuses had 5000 durability. If they were at full durability, they were nerfed to 2000.
```
My-Name-Is-Jeff, line 659: if > 2000 then 5000 else 2000 i guess would be the only way

not sure which is more preferable

lunaynx: That could kinda work, but it would suddenly jump to showing full durability once it reaches 2000.

PR #505, opened 2024-09-27
fix: Item Stars Display shows incorrect star count

TakoTheDev:


src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/item/ItemStackHook.kt
```diff
@@ -27,9 +27,10 @@ import gg.skytils.skytilsmod.utils.ifNull
 import net.minecraft.item.ItemStack
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable
 
-const val starPattern = "§6✪"
-val masterStars = ('➊'..'➎').toList()
-val masterStarPattern = Regex("§c[${masterStars.joinToString("")}]")
+const val star = "✪"
+val masterStars = ('➊'..'➎').map { it.toString() }
```
Sychic, line 31: what's the point of converting these to strings?

TakoTheDev: I can't use indexOf with a String on a Char list

TakoTheDev: And I think that the code using destructured is overall better than with the it.value.last

Sychic (approved):

PR #507, opened 2024-09-28
Added smolSelf secret setting

Challenge2you:
I (along with some other people I know) have gotten too used to the comfort of the smol people setting when playing in f5 mode mainly for hyperion movement.
What always bugged me was that some mob models also became small with this setting, making it annoying to spot them quickly and fight them, so I made this setting to allow players to have only themselves as a small model while everyone else stays big.

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
@@ -50,7 +50,7 @@ class EntityLivingBaseHook(val entity: EntityLivingBase) {
     }
 
     val isSmol by lazy {
-        Utils.inSkyblock && entity is EntityPlayer && (SuperSecretSettings.smolPeople || isBreefing)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
```diff
@@ -50,7 +50,7 @@ class EntityLivingBaseHook(val entity: EntityLivingBase) {
     }
 
     val isSmol by lazy {
-        Utils.inSkyblock && entity is EntityPlayer && (SuperSecretSettings.smolPeople || isBreefing)
+        Utils.inSkyblock && (entity is EntityPlayer && (SuperSecretSettings.smolPeople || isBreefing)) || (entity is EntityPlayerSP && SuperSecretSettings.smolSelf)
```
My-Name-Is-Jeff:
```suggestion
        Utils.inSkyblock && (entity is EntityPlayer && (SuperSecretSettings.smolPeople || (entity is EntityPlayerSP && SuperSecretSettings.smolSelf) || isBreefing))
```

This conditional should not be separated. An EntityPlayerSP is always an EntityPlayer

My-Name-Is-Jeff (changes requested):

PR #520, opened 2024-10-11
Fix slot number for Perkpocalypse perks in calendar

lunaynx:
Fixed Perkpocalypse perks not getting detected from the calendar due to the slot number being wrong.

Screenshot of calendar, showing mayor in slot 37


My-Name-Is-Jeff:
Is the calendar permanently like this now? Or is there a case where the row would be added/removed.

lunaynx:
I can't think of a case where it would be different, I remember the existing slot number was also wrong last Jerry when I made a PR to fix the condition being a `||` which ended up inadvertently breaking the check that happened to work before. Can't say for sure it was the exact same slot number then, but I *think* the calendar looked the same.

PR #523, opened 2024-12-12
feat: use mod API for ping packet

lunaynx:
Switches from using C16PacketClientStatus to ServerboundPingPacket from the Hypixel Mod API for sending pings.

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/core/Config.kt
@@ -2619,7 +2619,7 @@ object Config : Vigilant(
 
     @Property(
         type = PropertyType.SELECTOR, name = "Ping Display",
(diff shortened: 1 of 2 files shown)


lunaynx:
Kinda works, but not very well yet. I pretty much always get "Already pinging!" on manual pings and I seem to be getting FPS drops(?)

lunaynx:
So, I've fixed it, but the mod API seems to have a rate limit of 5 seconds on ping packets.

This works well enough for the ping display, not sure what to do about manual pings though, probably just defer to the last value if a ping was sent less than 5s ago.

lunaynx:
This seems to be causing occasional crashes, which appear to come from the Hypixel Mod API library, but I'm not 100% sure. I'll investigate further later.

lunaynx:
> As a side note, I'd recommend not spamming the ping packet. This packet is specifically implemented just for testing the basic idea of the API, and shouldn't be used to continuously poll the server.

(https://github.com/HypixelDev/ForgeModAPI/issues/10#issuecomment-2769219066)

PR #524, opened 2025-01-09
feat: option to hide damage splash

davishi25:


davishi25:
this feature adds an option to hide damage splash under the 'custom damage splash' drop-down in the same way that 'hide damage in boss' works.

PR #528, opened 2025-02-28
feat: added coleweight and soopy waypoint format

TakoTheDev:
from my limited testing it works.

Also this does not work with the weird soopy one.

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
@@ -29,6 +29,7 @@ import gg.skytils.skytilsmod.core.tickTimer
 import gg.skytils.skytilsmod.events.impl.skyblock.LocationChangeEvent
 import gg.skytils.skytilsmod.tweaker.DependencyLoader
 import gg.skytils.skytilsmod.utils.*
+import gg.skytils.skytilsmod.utils.rendering.addAll
 import kotlinx.serialization.EncodeDefault
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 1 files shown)


src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
```diff
@@ -370,3 +383,32 @@ data class Waypoint @OptIn(ExperimentalSerializationApi::class) constructor(
...
+    val x: Int,
+    val y: Int,
+    val z: Int,
+    val r: Int,
+    val g: Int,
+    val b: Int,
+    val options: ColeweightOptions
+) {
+    fun toSkytilsWaypoint(): Waypoint {
+        val color = Color(r * 255, g * 255, b * 255) // We assume that 0 is 0 and 1 is 255
```
My-Name-Is-Jeff: this doesn't really make sense to me if r, g, b are ints

src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
```diff
@@ -370,3 +383,32 @@ data class Waypoint @OptIn(ExperimentalSerializationApi::class) constructor(
...
+            true,
+            color,
+            System.currentTimeMillis(),
+            SkyblockIsland.CrystalHollows // We assume Crystal Hollows
+        )
+    }
+}
+
+@Serializable
+data class ColeweightOptions constructor(val name: String, )
```
My-Name-Is-Jeff: Make this a subclass

src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
```diff
@@ -370,3 +383,32 @@ data class Waypoint @OptIn(ExperimentalSerializationApi::class) constructor(
         UGraphics.enableDepth()
     }
 }
+
+@Serializable
+data class Coleweight(
```
My-Name-Is-Jeff: Maybe treat these like generic waypoints? have default values for r, g, b so that with just x, y, z it'll be accepted

My-Name-Is-Jeff (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
TakoTheDev: should the options (basically only name) also have a default value?

src/main/kotlin/gg/skytils/skytilsmod/features/impl/handlers/Waypoints.kt
TakoTheDev: ig JSON.stringify parses 0.0 and 1.0 to 0 and 1 respectively? Looking at their code they are supposed to be floats ig (in case I am wrong I added a coerceIn)

My-Name-Is-Jeff: > ig JSON.stringify parses 0.0 and 1.0 to 0 and 1 respectively? Looking at their code they are supposed to be floats ig (in case I am wrong I added a coerceIn)

In JavaScript all numbers are basically `double` types

PR #534, opened 2025-03-21
fix: totempos npe

Kemmisch:
totempos is nullable
not tested but should work, couldn't easily reproduce error


Kemmisch:
duplicates https://github.com/Skytils/SkytilsMod/pull/485

PR #541, opened 2025-05-06
Added a working smolMe option and made smolpeople work on players only

benjilmao:
I was bored and went to learn some kotlin, figured I'd try doing this since no one wants to based on dc messages.
added the secret settings "smolme" that shrinks the player without affecting other players, also added a check for stuff that use playerEntity to exclude them from being shrunk.
dont ask me to explain it more cuz idk how I even figured that out


src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
```diff
@@ -43,14 +43,20 @@ class EntityLivingBaseHook(val entity: EntityLivingBase) {
             overrideDisplayName = s
     }
 
+    private fun isNPC(): Boolean {
+        if (entity !is EntityPlayer) return true
+
+        return entity.hasCustomName() ||
+                entity.displayName.unformattedText.contains("♲") ||
```
Sychic: Why is this check needed?

src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
```diff
@@ -43,14 +43,20 @@ class EntityLivingBaseHook(val entity: EntityLivingBase) {
...
     val isBreefing by lazy {
-        entity.name == "Breefing" && (SuperSecretSettings.breefingDog || Random.nextInt(
-            100
-        ) < 3)
+        entity.name == "Breefing" && (SuperSecretSettings.breefingDog || Random.nextInt(100) < 3)
     }
 
     val isSmol by lazy {
-        Utils.inSkyblock && entity is EntityPlayer && (SuperSecretSettings.smolPeople || isBreefing)
+        Utils.inSkyblock && entity is EntityPlayer && (SuperSecretSettings.smolPeople || (entity is EntityPlayerSP && SuperSecretSettings.smolMe) || isBreefing ) && !isNPC()
```
Sychic: My first guess is that the npc check is only needed when going down the smolme path so I'd rather have it moved there to slightly optimize the condition check a little.

src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
benjilmao: My first thought while reading a doc was to have this check as a safety net to make sure smolpeople excludes npcs and mobs

src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
benjilmao: Looking at it now I could move it to smolpeople since EntityPlayerSP already excludes mobs and for smolme but I’ll have to test that when am home

src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/entity/EntityLivingBaseHook.kt
Sychic: No I mean that specific line looking for the recycling symbol.

benjilmao: That’s supposed to be “§” because Ive seen somewhere that hypixel uses formatting codes on mob names, idk why is changes that to the recycling icon though

Sychic: I'm not sure that works how you intend it to, players have colored names as well.

My-Name-Is-Jeff:
Duplicate of #507

PR #543, opened 2025-05-11
Improve performance of SmartFontRenderer.drawString

TheDGOfficial:
~~Fixes string width in SmartFontRenderer#drawString.~~ Improves performance in SmartFontRenderer#drawString.

The Regex is supposed to clean out the control codes in the format of &\<code\>, to get the raw string that's actually going to be rendered and use that for getStringWidth calculation.

However, the Regex was wrong and only matches the supposedly RGB codes like &[255,  47, 50], I don't understand what is the point in that so its likely just wrong Regex.

It was also appearing in Spark reports since its used in rendering code, if you had uncapped FPS limit (no VSync), the .toRegex() in Kotlin creates a new Regex every time drawString is called, which is equivalent to Pattern.compile in Java, which is not a free operation in terms of performance.

~~This commit just replaces the regex entirely with String#stripControlCodes, which is a Kotlin extension method provided by Utils.kt in Skytils and is used everywhere else in Skytils for removing control codes.~~

This PR just removes the Regex entirely, because as per my last comment (see https://github.com/Skytils/SkytilsMod/pull/543#issuecomment-2869830290), FontRenderer already handles 0 width color and formatting characters correctly and has special handling for bold formatting code giving extra width to the string, so removing the color codes actually would give incorrect string width.

<details>
<summary>Note for future (not relevant anymore)</summary>
Note: String#stripControlCodes calls UTextComponent.stripFormatting which calls EnumChatFormatting.getTextWithoutFormattingCodes, which still uses Regex to do replacing, but it uses the correct color code pattern and precompiles the pattern in a private static final Pattern field, so it does not call Pattern.compile each time, it just does formattingCodePattern.matcher(text).replaceAll("")

A highly performant version of color code removing without using Regex at all via StringBuilder is implemented [here](https://github.com/hannibal002/SkyHanni/blob/762bd80f1e267571081741283742b42e7d604085/src/main/java/at/hannibal2/skyhanni/utils/StringUtils.kt#L66) in SkyHanni and that can be copied if necessary in the future for even more performance, but that is out of the scope of this PR.
</details>

~~It should both fix the string width being larger than normal due to control codes and improve performance at the same time.~~
It should improve performance.

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/utils/graphics/SmartFontRenderer.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/utils/graphics/SmartFontRenderer.kt
@@ -20,6 +20,7 @@ package gg.skytils.skytilsmod.utils.graphics
 import gg.skytils.skytilsmod.utils.graphics.colors.CommonColors
 import gg.skytils.skytilsmod.utils.graphics.colors.CustomColor
 import gg.skytils.skytilsmod.utils.graphics.colors.MinecraftChatColors
+import gg.skytils.skytilsmod.utils.stripControlCodes
 import net.minecraft.client.Minecraft
 import net.minecraft.client.gui.FontRenderer
 import net.minecraft.client.renderer.GlStateManager
@@ -54,7 +55,7 @@ class SmartFontRenderer : FontRenderer(
         } else if (customColor === CommonColors.CRITICAL) {
             return drawCritText(text, x, y, alignment, shadow)
         }
-        val drawnText = text.replace("§\\[\\d+\\.?\\d*,\\d+\\.?\\d*,\\d+\\.?\\d*]".toRegex(), "")
+        val drawnText = text.stripControlCodes()
         return when (alignment) {
             TextAlignment.MIDDLE -> drawString(
                 text,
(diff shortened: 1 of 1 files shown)


TheDGOfficial:
After further checking, FontRenderer#getStringWidth seems to already take care of the fact that color codes add 0 width, and my changes actually cause another regression because FontRenderer#getStringWidth depending on if the text contains &l (bold) adds additional width. Removing all the color codes obviously makes font renderer never detect this bold parameter.

That being said, since FontRenderer#getStringWidth already takes into account color and formatting codes including special handling for bold, I'm even more confused why the original Regex and replace was there, in the first place, at all, so I'm going to just remove it all and pass the text directly to getStringWidth.

TheDGOfficial:
I wanted to rename the branch in GitHub and that seems to just delete the branch and create a new copy, which closed the PR lol. I'll see if I can change it without the PR closing.

TheDGOfficial:
Fine, i'll keep the branch name fix-string-width, I wanted to make it improve-performance-of-smartfontrenderer-drawstring since the PR no longer fixes width (since fontrenderer already handles that turns out) and only improves performance, but I couldnt really bother to deal with GitHub or Git bullshit at the moment and stackoverflow is not helping, I first renamed the branch on GitHub web interface which closed the PR, then renamed on local git and pushed and that still closed the PR after deleting the old branch. So just merge it like this I guess the branch name does not matter.

Sorry that it was a mess lol.

My-Name-Is-Jeff:
> I'm even more confused why the original Regex and replace was there, in the first place, at all, so I'm going to just remove it all and pass the text directly to getStringWidth.

I think it's so you could do `§[R,G,B]` but `drawChars` doesn't handle it, so I agree, it should be removed. Thanks for looking into this.

PR #544, opened 2025-05-11
Improve performance of SlayerFeatures.onRenderLivingPre

TheDGOfficial:
Avoids instanceof, cast, hasCustomName, getDisplayName, getUnformattedText calls if the feature was not enabled in the first place.

Not sure if any neglible performance gains (there is some because calling hasCustomName or getting displayName calls DataWatcher which has ReentrantReadWriteLock), but it would be bad practice to call methods and assign variables just to do a late if check and exit the method. Moving the config check up after inSkyblock seems like a better practice here.


TheDGOfficial:
Seems like we can also move the check for !mc.renderManager.isDebugBoundingBox up. Didn't see that at the first glance.

PR #548, opened 2025-05-27
fix: wrong entity error spam

TheDGOfficial:
When doing slayers, my logs were filled with thousands of Wrong entity! errors, bloating my logs folder and log sizes and making it harder to find real important exceptions. I asked about this by creating a [discussion](https://discord.com/channels/807302538558308352/1323888819229036585) in Skytils Discord which did not receive much input but we could not find the root cause of the error in the end (though I did close to zero debugging as just removing the exception seemed to fix the errors without breaking any slayer features) and I had the changes in my local Skytils build for months and had no issues, so I decided to finally upstream the changes for everyone to benefit.

Note that despite the error, all slayer features kept working fine.

There's already a printDevMessage above the original exception, so removing the exception does not cost anything in terms of debugging. A simple search of what the printDevMessage prints would already get us to the exact line in this file. Keep in mind that creating exceptions are not free in terms of performance and they have to generate a stack trace in addition to the exception object itself.

detectSlayerEntities was modified to be able return null and usages of detectSlayerEntities were modified to handle null return value gracefully to remove the throwing of the exception. Another alternative is to return a pair of 2 null values but creating unnecessary objects should be avoided.

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/features/impl/slayer/SlayerFeatures.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/features/impl/slayer/SlayerFeatures.kt
@@ -53,7 +53,7 @@ import gg.skytils.skytilsmod.utils.graphics.colors.CommonColors
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
-import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.launch
 import net.minecraft.block.BlockBeacon
 import net.minecraft.client.entity.EntityOtherPlayerMP
@@ -339,9 +339,10 @@ object SlayerFeatures : CoroutineScope {
             ) {
                 slayer?.run {
                     launch {
-                        val (n, t) = detectSlayerEntities().first()
-                        nameEntity = n
-                        timerEntity = t
+                        detectSlayerEntities().firstOrNull()?.let { (n, t) ->
+                            nameEntity = n
+                            timerEntity = t
+                        }
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/gg/skytils/skytilsmod/features/impl/slayer/SlayerFeatures.kt
```diff
@@ -53,7 +53,7 @@ import gg.skytils.skytilsmod.utils.graphics.colors.CommonColors
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
-import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
```
My-Name-Is-Jeff, line 56: Not sure if this is the correct function for us to use here, `first` should still work? Don't know if changing the method might cause any side effects.

Sychic:
This pr hides the actual cause of the problem where the entity is not detected. Instead this simply hides the resulting symptom of errors being spammed.

TheDGOfficial:
> This pr hides the actual cause of the problem where the entity is not detected. Instead this simply hides the resulting symptom of errors being spammed.

I have since then found the cause. The error appears to happen when theres no boss for a second (not sure how many ticks) when demons are killed before the boss respawns. This for Blaze T4 but I remember getting this error when doing Enderman slayer as well before, though I'm not sure about the root cause on that, as far as I know lazer phase does not despawn the enderman boss, so maybe something else going on there.

Regarding the "this hides the resulting symptom of errors" part, The errors are already only printed to logs with a useless long stacktrace (stacktrace is useless here because theres already a debug call above, which I did not remove in the pr intentionally). The usual "Skytils caught a X error, please report on Discord" line does not appear in chat (if it did I'm sure the support channel would've flooded with it). It's a silent error (so it was already kind of hidden - i am not hiding anything more in the pr) that results in big log files after sessions of doing slayers.

I'm just going to keep using this solution in my local ST build but feel free to fix it upstream. I'm not sure if it can be fixed any other way than removing the exception - maybe setting a grace period where no errors are thrown for like 20 ticks after demons are killed and keep retrying searching the boss and only error out 1 time if it wasn't found after that? In my opinion no need to complicate things - just not throwing an exception is the better solution. At worst case it can be a single log message instead of a long exception. It really bloats the log files, especially after long slayer sessions.

Sychic:
In theory, the actual blaze entity never disappears. I was going to rewrite slayer handling but I've since forgotten how I planned on doing this. I think one possible way of doing this is instead of waiting a tick and blocking threads, instead handle things as soon as we have info for them on the client (via packets) but this would likely be ruined because of how hypixel's nametags aren't easily correlated with entity. I'm open to suggestions on how to do a rewrite but I would still like to be able to tell when the entity detection fails. As for this PR, I'm fine with reopening if you instead change it to some other form of logging, a simple line should suffice.

PR #564, opened 2025-06-25
feat(2.x): Use Modrinth for Update Checking

My-Name-Is-Jeff:


mod/build.gradle.kts
```diff
@@ -274,7 +274,7 @@ tasks {
         enabled = false
     }
     named<RemapJarTask>("remapJar") {
-        archiveBaseName.set("Skytils")
+        archiveBaseName.set(shadowJar.flatMap { it.archiveBaseName })
```
Sychic:
```suggestion
        archiveBaseName.set(shadowJar.flatMap(ShadowJar::archiveBaseName))
```
I think this works but idk what the type is so hopefully this works

PR #570, opened 2025-07-07
fix: sodium block state spoof

P3nguinMinecraft:
Fixes incompatibility with **Sodium Fabric** because it uses its own chunk render pipeline, breaking **Dark Most Mist** and **Recolor Dwarven Carpets** (never reaches SectionBuilder mixin)

- ~~Added sodium jar as modCompileOnly dependency under directory `mod/1.21.5-fabric/libs`~~
  - ~~You can add sodium as submodule and add to compile but I dont wanna mess with that~~
- ~~`libs` is included in build script under flatDir with fabric platform condition~~
- Uses mixin to `net.caffeinemc.mods.sodium.client.world.LevelSlice`, redirected to `BlockRendererDispatcherHook`
- DispatcherHook is slightly rewritten to handle `SectionBuilder` and `LevelSlice` mixin arg types
- ~~Sodium mixin is conditionally added through `SkytilsMixinPlugin`, which detects sodium client class~~

### DIFF
--- a/mod/src/main/java/gg/skytils/skytilsmod/mixins/transformers/renderer/sodium/MixinLevelSlice.java
+++ b/mod/src/main/java/gg/skytils/skytilsmod/mixins/transformers/renderer/sodium/MixinLevelSlice.java
@@ -18,22 +18,22 @@
 
 package gg.skytils.skytilsmod.mixins.transformers.renderer.sodium;
 
+import com.llamalad7.mixinextras.injector.ModifyReturnValue;
+import com.llamalad7.mixinextras.sugar.Local;
 import gg.skytils.skytilsmod.mixins.hooks.renderer.BlockRendererDispatcherHookKt;
 import net.minecraft.block.BlockState;
+import net.minecraft.util.math.BlockPos;
+import org.spongepowered.asm.mixin.Dynamic;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Pseudo;
 import org.spongepowered.asm.mixin.injection.At;
-import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
 @Pseudo
 @Mixin(targets = "net.caffeinemc.mods.sodium.client.world.LevelSlice")
 public class MixinLevelSlice {
-    @Inject(method = "getBlockState(III)Lnet/minecraft/block/BlockState;", at = @At("RETURN"), cancellable = true)
-    private void modifyBlockState(int blockX, int blockY, int blockZ, CallbackInfoReturnable<BlockState> cir){
-        BlockState state = BlockRendererDispatcherHookKt.modifyBlockState(blockX, blockY, blockZ);
-        if (!cir.getReturnValue().equals(state)){
-         cir.setReturnValue(state);
-        }
+    @Dynamic
+    @ModifyReturnValue(method = "getBlockState(III)Lnet/minecraft/block/BlockState;", at = @At("RETURN"))
+    private BlockState modifyBlockState(BlockState original, @Local(argsOnly = true) int blockX, @Local(argsOnly = true) int blockY, @Local(argsOnly = true) int blockZ){
+        return BlockRendererDispatcherHookKt.modifyBlockState(new BlockPos(blockX, blockY, blockZ), original);
     }
 }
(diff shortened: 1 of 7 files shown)


mod/src/main/java/gg/skytils/skytilsmod/mixins/transformers/renderer/sodium/MixinLevelSlice.java
```diff
@@ -0,0 +1,35 @@
...
+
+import gg.skytils.skytilsmod.mixins.hooks.renderer.BlockRendererDispatcherHookKt;
+import net.caffeinemc.mods.sodium.client.world.LevelSlice;
+import net.minecraft.block.BlockState;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(LevelSlice.class)
```
My-Name-Is-Jeff: You can do targets = "net.caf....LevelSlice" and then add the Pseudo annotation

mod/src/main/java/gg/skytils/skytilsmod/mixins/transformers/renderer/sodium/MixinLevelSlice.java
```diff
@@ -0,0 +1,35 @@
...
+public class MixinLevelSlice {
+    @Inject(method = "getBlockState(III)Lnet/minecraft/block/BlockState;", at = @At("RETURN"), cancellable = true)
+    private void modifyBlockState(int blockX, int blockY, int blockZ, CallbackInfoReturnable<BlockState> cir){
+        cir.setReturnValue(BlockRendererDispatcherHookKt.modifyBlockState(blockX, blockY, blockZ));
```
My-Name-Is-Jeff: Use ModifyReturnValue injector

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/BlockRendererDispatcherHook.kt
```diff
@@ -24,16 +24,29 @@ import gg.skytils.skytilsmod.utils.SkyblockIsland
...
     instance: ChunkRendererRegion,
     pos: BlockPos,
     original: BlockState
 ): BlockState {
+    return modifyBlockState(pos, original)
+}
+
+fun modifyBlockState(blockX: Int, blockY: Int, blockZ: Int): BlockState {
+    val pos = BlockPos(blockX, blockY, blockZ)
+    val original = client.world?.getBlockState(pos) ?: Blocks.AIR.defaultState
```
My-Name-Is-Jeff: this isn't necessary, on an inject you can use cir.getReturnValue()

ModifyReturnValue will give you the original

P3nguinMinecraft:
i need to reposition libs lol

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/BlockRendererDispatcherHook.kt
P3nguinMinecraft: you need the y value

My-Name-Is-Jeff: Use @Local

My-Name-Is-Jeff: It might actually be better to keep it as Inject but don't always call setReturnValue

P3nguinMinecraft: so is this good
``` java
@Inject(method = "getBlockState(III)Lnet/minecraft/block/BlockState;", at = @At("RETURN"), cancellable = true)
  private void modifyBlockState(int blockX, int blockY, int blockZ, CallbackInfoReturnable<BlockState> cir){
      BlockState state = BlockRendererDispatcherHookKt.modifyBlockState(blockX, blockY, blockZ);
      if (cir.getReturnValue().equals(state)){
          cir.setReturnValue(state);
      }
  }
```

My-Name-Is-Jeff (approved):

PR #571, opened 2025-07-07
fix: smolpeople breefingdog player sizing

P3nguinMinecraft:
- spoofing isBaby() does not work anymore
- changing player size for smolpeople and breefingdog will be done in PlayerEntityRenderer mixin, scaling the MatrixStack by 0.5f
- another issue is `by lazy` was not going to work because all players would share the same hook instance, so I just made the smol / breefing check per call

I am not quite sure if I matched the extension / hook system well


P3nguinMinecraft:
im actually so bad at branches bruh accidentally added null world age fix in here somehow

CalMWolfs:
Wouldn't these changes now break smolpeople on 1.8. Due to the presenece of preprocessing in the repo id assume there are still 1.8 releases planned so instead of removing the old code you would prob have to preprocess it out instead

Sychic:
I can fix the commits in a bit.

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/extensions/ExtensionPlayerEntityRenderer.kt
Sychic, line 1: Is there a reason for this ext interface? afaict, nowhere accesses this value and as such it shouldn't need to be stored.

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/PlayerEntityRendererHook.kt
Sychic, line 1: These functions can just be declared in the file itself and referenced from java using the filekt format. i.e:
```kt
package gg.skytils.skytilsmod.mixins.hooks.renderer

fun isBreefing()
```
```java
PlayerEntityRendererHookKt.isBreefing()
```

mod/src/main/java/gg/skytils/skytilsmod/mixins/transformers/renderer/MixinPlayerEntityRenderer.java
```diff
@@ -0,0 +1,47 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+//#if MC>12000
+package gg.skytils.skytilsmod.mixins.transformers.renderer;
```
Sychic, line 18: I believe you'd want to retain the package declaration

Sychic (changes requested):

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/extensions/ExtensionPlayerEntityRenderer.kt
P3nguinMinecraft: I was just trying to copy the structure for other hooks, ig its not needed

P3nguinMinecraft: if you want I can get rid of the hook and just handle everything within the mixin

Sychic: That would be preferred.

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/PlayerEntityRendererHook.kt
```diff
@@ -24,16 +24,14 @@ import net.minecraft.client.render.entity.state.PlayerEntityRenderState
...
         = state.name == "Breefing" && (SuperSecretSettings.breefingDog || Random.nextInt(
             100
         ) < 3)
 
-    fun smol(state: PlayerEntityRenderState, ms: MatrixStack) {
-        if (Utils.inSkyblock && (SuperSecretSettings.smolPeople || isBreefing(state))){
-            ms.scale(0.5f, 0.5f, 0.5f)
-        }
-    }
+    @JvmStatic
```
Sychic: You don't need this jvmstatic if you reference the INSTANCE of the class.

Sychic:
Is there a reason why the logic is now performed in the mixin itself? It would make more sense imo to simply retain the existing logic (lazy caches the value once computed).

P3nguinMinecraft:
> Is there a reason why the logic is now performed in the mixin itself? It would make more sense imo to simply retain the existing logic (lazy caches the value once computed).

caching the value was an issue because it kept computing before you loaded in or smth and you would have to change lobbies to apply settings changes

Sychic:
Can you change the behavior so 1.8.9 works the same way then?

mod/src/main/kotlin/gg/skytils/skytilsmod/mixins/hooks/renderer/PlayerEntityRendererHook.kt
```diff
@@ -0,0 +1,38 @@
...
+import gg.skytils.skytilsmod.utils.Utils
+import net.minecraft.client.render.entity.state.PlayerEntityRenderState
+import net.minecraft.client.util.math.MatrixStack
+import kotlin.random.Random
+
+class PlayerEntityRendererHook {
+    fun isBreefing(state: PlayerEntityRenderState): Boolean
+        = state.name == "Breefing" && (SuperSecretSettings.breefingDog || Random.nextInt(
+            100
+        ) < 3)
```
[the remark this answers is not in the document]
P3nguinMinecraft: TODO: flickering for breefing

mod/src/main/java/gg/skytils/skytilsmod/mixins/transformers/entity/MixinEntityLivingBase.java
```diff
@@ -82,10 +82,13 @@ private boolean spawnParticle(World instance, Entity entity, byte b) {
     //#endif
     }
 
-    @Inject(method = "isBaby", at = @At("HEAD"), cancellable = true)
-    private void setChildState(CallbackInfoReturnable<Boolean> cir) {
-        hook.isChild(cir);
-    }
+    //#if MC>12000
+    //#else
```
Sychic:
```suggestion
    //#if MC==10809
```

Sychic (changes requested):
The logic for checking `isSmol` and `isBreefing` should be consolidated in the hook and reused across versions. looks like the only version specific thing is accessing the name which can just be supplied in a function invokation.

PR #573, opened 2025-07-10
fix: fov drifting (waypoint blocks)

P3nguinMinecraft:
fixed waypoint block offsetting because of fov not normalized (70)
it would affect and prob fix (or break) other renders that use `setupCameraTransformations()` but those do not work currently lol

couldnt figure out how to fix the nametags im just throwing rocks at a wall atp
only solution i have is just brute force calculating offset but that isnt great

### DIFF
--- a/mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
+++ b/mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
@@ -55,7 +55,7 @@ object DrawHelper {
      */
     fun cameraRotation(matrices: UMatrixStack) {
         //#if MC>12000
-        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()).mul(70f/mc.options.fov.value.coerceAtLeast(1))
+        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()).mul(70f/mc.options.fov.value.coerceAtLeast(1)))
         //#else
         //$$matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()))
         //#endif
--- a/mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
+++ b/mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
@@ -54,7 +54,7 @@ object DrawHelper {
      */
     fun cameraRotation(matrices: UMatrixStack) {
         //#if MC>12000
-        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()).mul(70f/mc.options.fov.value))
+        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()).mul(70f/mc.options.fov.value.coerceAtLeast(1))
         //#else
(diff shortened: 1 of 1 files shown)


mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
```diff
@@ -53,7 +53,11 @@ object DrawHelper {
      * This is useful for rendering things in world space, as it will negate the camera rotation.
      */
     fun cameraRotation(matrices: UMatrixStack) {
-        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()))
+        //#if MC>12000
+        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()).mul(70f/mc.options.fov.value))
```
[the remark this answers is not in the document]
P3nguinMinecraft: @gemini-code-assist
fov cannot be 0 unless someone is doing some really funny stuff

Sychic:
```suggestion
        matrices.multiply(mc.gameRenderer.camera.rotation.conjugate(Quaternionf()).mul(70f/mc.options.fov.value.coerceAtLeast(1))
```

mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
```diff
@@ -223,9 +227,9 @@ object DrawHelper {
     fun drawNametag(matrices: UMatrixStack, text: String, x: Double, y: Double, z: Double, shadow: Boolean = true, scale: Float = 1f, background: Boolean = true, throughWalls: Boolean = false) {
         matrices.push()
         matrices.translate(x, y + 0.5, z)
-        matrices.multiply(mc.entityRenderDispatcher.rotation)
+        matrices.multiply(mc.gameRenderer.camera.rotation)
```
Sychic: Is there a reason for this new value? seems like fov drifting for nametags wasn't fixed so is there a reason to use this new value?

Sychic (changes requested):

My-Name-Is-Jeff:
If the FOV is the cause for the rendering position issues, that would indicate that the projection matrix is incorrect? I feel like then we should aim to copy the MC projection matrix over but I'm not sure.

mod/src/main/kotlin/gg/skytils/skytilsmod/utils/rendering/DrawHelper.kt
P3nguinMinecraft: the value should be the same its just for consistency, i can change it back if it affects things

P3nguinMinecraft:
this should at least fix for all box renders like in dungeons, text will need a diff fix

The-Deerman:
~~Yeah on Fabric you don't need to cancel out the camera rotation, the built-in matrix stack for world rendering already has the camera rotation removed before your render callback runs. So when you also cancel it manually, you’re canceling it twice → things spin in the opposite direction. I just removed cameraRotation(matrices) from setupCameraTransformations and that stops the outlined bounding box from moving when looking around. Still moves when shifting~~

Yeah I don't know anymore

Sychic (changes requested):
This feels like a bandaid fix for an underlying issue. While it may work, it feels like this will eventually become an issue again (it also doesn't fix nametags)

PR #581, opened 2025-08-04
fix: getNbt

P3nguinMinecraft:
explicitly returns null if trying to get nbt of empty itemstack, which toNbt used to throw

### DIFF
--- a/mod/src/main/kotlin/gg/skytils/skytilsmod/utils/multiplatform/nbt.kt
+++ b/mod/src/main/kotlin/gg/skytils/skytilsmod/utils/multiplatform/nbt.kt
@@ -24,5 +24,5 @@ import kotlin.jvm.optionals.getOrNull
 
 val ItemStack.nbt
     get() =
-        if (this.isEmpty) null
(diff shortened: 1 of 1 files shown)


mod/src/main/kotlin/gg/skytils/skytilsmod/utils/multiplatform/nbt.kt
```diff
@@ -25,4 +25,4 @@ import kotlin.jvm.optionals.getOrNull
 val ItemStack.nbt
     get() =
         if (this.isEmpty) null
-        else MinecraftClient.getInstance().player?.registryManager?.let(::toNbt)?.asCompound()?.getOrNull()
\ No newline at end of file
+        else MinecraftClient.getInstance().player?.registryManager?.let(this::toNbt)?.asCompound()?.getOrNull()
```
Sychic: If you're going to qualify this, I'd prefer also qualifying `this@ItemStack`

Sychic (approved):

PR #582, opened 2025-08-05
fix: make Hide Non-Starred Mob Nametags work with new mob type update

DaysSky:
new alpha update adds non-letter characters before "✯ "

### DIFF
--- a/mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/DungeonFeatures.kt
+++ b/mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/DungeonFeatures.kt
@@ -476,7 +476,7 @@ object DungeonFeatures : EventSubscriber {
                     if (name.contains("Terracotta ")) mc.world?.removeEntity(event.entity.id, Entity.RemovalReason.DISCARDED)
                 }
                 if (Skytils.config.hideNonStarredNametags) {
-                    if (!name.startsWith("✯ ") && name.contains("❤") && dungeonMobSpawns.any { it in name }) {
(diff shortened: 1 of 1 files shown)


mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/dungeons/DungeonFeatures.kt
```diff
@@ -476,7 +476,7 @@ object DungeonFeatures : EventSubscriber {
                     if (name.contains("Terracotta ")) mc.world?.removeEntity(event.entity.id, Entity.RemovalReason.DISCARDED)
                 }
                 if (Skytils.config.hideNonStarredNametags) {
-                    if (!name.startsWith("✯ ") && name.contains("❤") && dungeonMobSpawns.any { it in name }) {
+                    if (!name.takeWhile{!it.isLetter()}.contains("✯ ") && name.contains("❤") && dungeonMobSpawns.any { it in name }) {
```
Sychic: I'm hesitant to do it this way, if the characters are always the first section before the first space, we can just substring before the first space and check that.

DaysSky: there is a space between mob types and the star

Sychic: could we simply check for the existence of a star at all in the name?

DaysSky: i thought that would be slower but turns out its not

Sychic (approved):

PR #584, opened 2025-09-20
Slayer and miniboss detection

The-Deerman:
Fix: Detection for minibosses by removing the color codes which aren't present in entity.displayName?.string
Fix: Slayer detection by using contains instead of startsWith so that even if hypixel changes their mobtypes again it won't instantly break

Doesn't fix any of the rendering issues

### DIFF
--- a/mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/slayer/SlayerFeatures.kt
+++ b/mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/slayer/SlayerFeatures.kt
@@ -82,14 +82,14 @@ object SlayerFeatures : EventSubscriber, CoroutineScope {
     override val coroutineContext = Executors.newSingleThreadExecutor().asCoroutineDispatcher() + SupervisorJob()
 
     private val ZOMBIE_MINIBOSSES = arrayOf(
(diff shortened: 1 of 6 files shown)


mod/src/main/kotlin/gg/skytils/skytilsmod/features/impl/slayer/SlayerFeatures.kt
```diff
@@ -256,7 +256,7 @@ object SlayerFeatures : EventSubscriber, CoroutineScope {
             val entity = event.entity as ArmorStandEntity
             if (!entity.hasCustomName()) return
             val name = entity.displayName?.string ?: return
-            if (Skytils.config.slayerBossHitbox && name.endsWith("§c❤") && !name.endsWith("§e0§c❤") && !mc.entityRenderDispatcher.shouldRenderHitboxes()) {
+            if (Skytils.config.slayerBossHitbox && name.endsWith("❤") && !name.endsWith(" 0❤") && !mc.entityRenderDispatcher.shouldRenderHitboxes()) {
```
Sychic, line 259: Are you sure this works? Assuming that name is still formatted, this will fail due to having no formatting codes

Sychic (commented):
Overall looks fine, just one question

PR #588, opened 2026-02-21
feat: item stars display

TakoTheDev:
while it works, I don't think Injecting ItemStack#getName is the most performant idea (may be better to just do it specifically for HandledScreen Tooltip Rendering and InGameHud Item Name Popup)

image
image

### DIFF
--- a/src/main/kotlin/gg/skytils/skytilsmod/hooks/ItemStackHook.kt
+++ b/src/main/kotlin/gg/skytils/skytilsmod/hooks/ItemStackHook.kt
@@ -1,0 +1,61 @@
+package gg.skytils.skytilsmod.hooks
+
+import gg.essential.universal.utils.toUnformattedString
+import gg.skytils.skytilsmod.core.Config
+import gg.skytils.skytilsmod.util.SBInfo
+import net.minecraft.text.Text
+import net.minecraft.util.Formatting
+
+const val star = '✪'
+val masterStars = ('➊'..'➎').map { it.toString() }
+
+fun modifyDisplayName(displayName: Text): Text {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/gg/skytils/skytilsmod/hooks/ItemStackHook.kt
Sychic, line 1: In general, I'd like to avoid this kind of hook style class. I think it's fine if you want a kotlin helper to handle parsing the item name but imo business logic should be directly in the mixin.

src/main/kotlin/gg/skytils/skytilsmod/hooks/ItemStackHook.kt
```diff
@@ -0,0 +1,61 @@
...
+
+            // 1 = Old, 2 = Compact, 0 = Disabled
+            when (Config.starDisplayType) {
+                1 -> {
+                    out.append(current)
+
+                    if (masters > 0) {
+                        out.append(Text.literal(star.toString().repeat(masters)).formatted(Formatting.RED))
+                    }
+                }
```
Sychic, line 38: Did it really use to just put the stars on at the end? I remember it would replace the stars. i.e. if it was 8 starts it'd be 3 red and 2 gold.

Sychic (changes requested):

src/main/kotlin/gg/skytils/skytilsmod/hooks/ItemStackHook.kt
TakoTheDev: lmao yea I remembered it wrong, you're right

will fix tomorrow probably

